PR #7, opened 2023-08-25
add typed assignment functions to support dynamic configuration (FF-803)

leoromanovsky:
## motivation

Currently always returning a `String` for `getAssignment`

## description

* adds new typed `get*Assignment` methods for `bool`, `double` and `string`
* enables github unit tests by creating a `testTarget`
* delete test data stored in git; keeps consistent with other repos
* version bump to 1.0.0


evanmiller (approved):

.github/workflows/unit-tests.yml
```diff
@@ -1,15 +1,23 @@
...
+name: Swift
 
-# on: [push]
+on: [push]
 
-# jobs:
-#   build:
+jobs:
+  build:
+    runs-on: macos-latest
```
aarsilv, line 7: 🤯 
I remember when macos was a no-go for the cloud and we had a stack of mac minis in the office with a sticky note on their UPS saying DO NOT UNPLUG

Sources/eppo/EppoClient.swift
```diff
@@ -45,14 +45,59 @@ public class EppoClient {
...
+        return try getAssignment(subjectKey, flagKey, SubjectAttributes())?.boolValue()
+    }
+    
+    public func getBoolAssignment(
+        _ subjectKey: String,
+        _ flagKey: String,
+        _ subjectAttributes: SubjectAttributes) throws -> Bool?
+    {
+        return try getAssignment(subjectKey, flagKey, subjectAttributes)?.boolValue()
+    }
```
aarsilv, line 61: I think we can collapse these into one function declaration [leveraging default parameter values](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/functions#Default-Parameter-Values).

```swift
public func getBoolAssignment(
    _ subjectKey: String,
    _ flagKey: String,
    _ subjectAttributes: SubjectAttributes = SubjectAttributes()) throws -> Bool?
{
    return try getAssignment(subjectKey, flagKey, subjectAttributes)?.boolValue()
}
```

Same for the other types below

Sources/eppo/EppoClient.swift
```diff
@@ -45,14 +45,59 @@ public class EppoClient {
...
+    {
+        return try getAssignment(subjectKey, flagKey, subjectAttributes)?.stringValue()
+    }
+    
+    private func getAssignment(_ subjectKey: String, _ flagKey: String) throws -> EppoValue? {
         return try getAssignment(subjectKey, flagKey, [:]);
     }
 
-    public func getAssignment(
+    private func getAssignment(
```
aarsilv, line 97: By making this private, could we be breaking things for SDK users not explicitly using the typed value methods yet?

.github/workflows/unit-tests.yml
leoromanovsky: Now it's an enclosure around a Mac mini 🤣

Sources/eppo/EppoClient.swift
leoromanovsky: Yes this is a breaking change and we will set the version number to a "major" bump.

PR #8, opened 2023-09-07
Add test data from github; rename json response to getJSONStringAssig… …nment (FF-890, FF-900)

leoromanovsky:
**api changes**

* Add back `getAssignment` for backwards compatibility
* Add `getJSONStringAssignment` that returns a `String`

**implementation details**

We are returning stringified JSON - in order to simplify the implementation we do not need to parse the string into swift json and back to a string. Instead we are supporting both `value` and `typedValue` fields in `Variation` - in the case of requesting a JSON response we read and pass-through the stringified JSON.

https://www.loom.com/share/6f2dc59f9192487486da11129f97f461

**house-keeping**

* add test data files from github to be able to share across SDK repos

**testing**

all unit tests pass

### DIFF
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -135,6 +135,7 @@ public class EppoClient {
             return nil;
         }
 
+        // Access the stringified JSON from `value` field until support for native JSON is available.
         if (isJson) {
             return EppoValue(value: assignedVariation.value, type: EppoValueType.String);
         }
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -50,7 +50,7 @@ public class EppoClient {
         _ flagKey: String,
         _ subjectAttributes: SubjectAttributes) throws -> String?
     {
-        return try getStringAssignment(subjectKey, flagKey, subjectAttributes)
+        return try getInternalAssignment(subjectKey, flagKey, subjectAttributes, false)?.stringValue()
     }
 
     public func getBoolAssignment(
@@ -58,7 +58,7 @@ public class EppoClient {
         _ flagKey: String,
         _ subjectAttributes: SubjectAttributes = SubjectAttributes()) throws -> Bool?
     {
-        return try getInternalAssignment(subjectKey, flagKey, subjectAttributes)?.boolValue()
+        return try getInternalAssignment(subjectKey, flagKey, subjectAttributes, true)?.boolValue()
     }
     
     public func getJSONStringAssignment(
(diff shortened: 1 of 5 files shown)


Sources/eppo/EppoClient.swift
```diff
@@ -45,39 +45,51 @@ public class EppoClient {
         self.flagConfigs = try JSONDecoder().decode(FlagConfigJSON.self, from: urlData);
     }
     
+    public func getAssignment(
+        _ subjectKey: String,
+        _ flagKey: String,
+        _ subjectAttributes: SubjectAttributes) throws -> String?
+    {
+        return try getStringAssignment(subjectKey, flagKey, subjectAttributes)
```
aarsilv, line 53: I don't think we can just call `getStringAssignment()` because while similar these two do different things:
* `getAssignment()` should just return the old-school, string version of the assignment like we were doing before these typed methods.
* `getStringAssingent()` should return the string assignment _if it is a string_.

So if the assignment is a numeric assignment (e.g., `123`) then `getAssignment()` should return `"123"`, `getNumericAssignment()` should return `123`, and `getStringAssginment()` should return `nil`.

Since we want `getAssignment()` to work regardless of type, we'll want something like a `toString()` on `EppoValue` (which we'll need for logging allocations anyways) and should return that.

(Alternatively, we could also store the untyped value somewhere and just pass that through)

leoromanovsky: Good point - I will return the contents of `variation.value` in this case.

aarsilv (approved):
Thanks for iterating! Approving as is as the typed getters should work as expected in the happy path and the untyped getter should continue to be backwards compatible.

That being said, there is a gap between what happens with this SDK and the "ideal" scenario in the edge case / error condition where a developer uses the wrong typed getter (e.g., calls `getNumericAssignment()` for a value that is a string type. Here we will return the casted version. In the Java (and soon) PHP versions we'll return null.

This discrepancy already existed before you got in here. However while in here, this could be a nice chance to fix it up.

leoromanovsky:
> Thanks for iterating! Approving as is as the typed getters should work as expected in the happy path and the untyped getter should continue to be backwards compatible.
> 
> That being said, there is a gap between what happens with this SDK and the "ideal" scenario in the edge case / error condition where a developer uses the wrong typed getter (e.g., calls `getNumericAssignment()` for a value that is a string type. Here we will return the casted version. In the Java (and soon) PHP versions we'll return null.
> 
> This discrepancy already existed before you got in here. However while in here, this could be a nice chance to fix it up.

Yea I see what you mean now - ok yea I understand and am up for further improvements.

PR #11, opened 2024-02-08
add assignment logger optional parameter and invocation (FF-998)

leoromanovsky:
A few considers to discuss:

* if there is an exception within the provided assignment logger should we recover
* should we try to invoke the assignment logger in a background thread to avoid blocking program flow


aarsilv (approved):
> if there is an exception within the provided assignment logger should we recover
Yeah, in other SDKs we log a warning and carry on (e.g., [JS SDK](https://github.com/Eppo-exp/js-client-sdk-common/blob/main/src/client/eppo-client.ts#L596))
> should we try to invoke the assignment logger in a background thread to avoid blocking program flow
Ideally

While both of these are _ideal_, I defer to you if you want to do it now or do it later.

PR #16, opened 2024-03-28
Add assignment cache (FF-1821)

leoromanovsky:
## motivation

The assignment cache in the javascript eco-system is useful for preventing duplicate assignment logging events and saving our customers' compute and storage cost.

* eppo javascript commons implementation - https://github.com/Eppo-exp/js-client-sdk-common/pull/25
* https://docs.geteppo.com/sdks/client-sdks/javascript#avoiding-duplicated-assignment-logs

## description

Client SDKs are all tied to a single subject - as such we can afford to implement a non-expiring cache.

Adds an interface of `AssignmentCache` and an implementation of `InMemoryAssignmentCache` using dictionaries. This will be useful for removing duplicate assignments **_within a user session_**. Further refinements can be added to persist this cache on disk to make it remove duplicates across sessions.

## testing

1. Unit tests for the cache class
2. Augmented eppo client tests with a focus on avoiding logging assignment logs. Verify that the cache key, and changes to the configuration, correctly log.

### DIFF
--- a/Sources/eppo/AssignmentCache.swift
+++ b/Sources/eppo/AssignmentCache.swift
@@ -1,0 +1,53 @@
+import Foundation
+
+public protocol AssignmentCache {
+    func hasLoggedAssignment(key: AssignmentCacheKey) -> Bool
+    func setLastLoggedAssignment(key: AssignmentCacheKey)
+}
+
+public struct AssignmentCacheKey {
+    var subjectKey: String
+    var flagKey: String
+    var allocationKey: String
+    var variationValue: EppoValue
+}
+
+public class InMemoryAssignmentCache: AssignmentCache {
+    private var cache: [String: String] = [:]
+
+    // This empty constructor is required to be able to instantiate the class
+    // within the constructor of EppoClient.
+    public init() {
+        // Initialization code here
+    }
+
+    public func hasLoggedAssignment(key: AssignmentCacheKey) -> Bool {
+        let cacheKey = getCacheKey(key: key)
+        if !has(key: cacheKey) {
+            return false
+        }
+
+        return get(key: cacheKey) == key.variationValue.toHashedString()
+    }
+
+    public func setLastLoggedAssignment(key: AssignmentCacheKey) {
+        let cacheKey = getCacheKey(key: key)
+        set(key: cacheKey, value: key.variationValue.toHashedString())
+    }
+
+    internal func get(key: String) -> String? {
+        return cache[key]
+    }
+
+    internal func set(key: String, value: String) {
+        cache[key] = value
+    }
+
+    internal func has(key: String) -> Bool {
+        return cache[key] != nil
+    }
+
+    private func getCacheKey(key: AssignmentCacheKey) -> String {
+        return ["subject:\(key.subjectKey)", "flag:\(key.flagKey)", "allocation:\(key.allocationKey)"].joined(separator: ";")
+    }
+}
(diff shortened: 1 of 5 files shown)


Tests/eppo/ClientTests.swift
```diff
@@ -20,13 +20,30 @@ class EppoMockHttpClient: EppoHttpClient {
...
+        self.jsonResponse = jsonResponse.data(using: .utf8)
+    }
+    
+    func get(_ url: URL) async throws -> (Data, URLResponse) {
+        guard let jsonData = jsonResponse else {
+            throw URLError(.badServerResponse)
+        }
+        return (jsonData, URLResponse(url: url, mimeType: "application/json", expectedContentLength: jsonData.count, textEncodingName: nil))
+    }
+}
```
leoromanovsky, line 36: directly provide a JSON string instead of loading from disk like the other mock - this is useful for changing the configuration within a test, as I do below.

Sources/eppo/EppoClient.swift
```diff
@@ -31,11 +30,13 @@ public class EppoClient {
     public init(
         _ apiKey: String,
         host: String = "https://fscdn.eppo.cloud",
-        assignmentLogger: AssignmentLogger? = nil
+        assignmentLogger: AssignmentLogger? = nil,
+        assignmentCache: AssignmentCache? = InMemoryAssignmentCache()
```
leoromanovsky, line 34: enables the cache by default - users can disable it or provide their own implementation

Sources/eppo/AssignmentCache.swift
```diff
@@ -0,0 +1,53 @@
...
+    var variationValue: EppoValue
+}
+
+public class InMemoryAssignmentCache: AssignmentCache {
+    private var cache: [String: String] = [:]
+
+    // This empty constructor is required to be able to instantiate the class
+    // within the constructor of EppoClient.
+    public init() {
+        // Initialization code here
```
schmit, line 21: can remove this comment?

Sources/eppo/AssignmentCache.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    internal func set(key: String, value: String) {
+        cache[key] = value
+    }
+
+    internal func has(key: String) -> Bool {
+        return cache[key] != nil
+    }
+
+    private func getCacheKey(key: AssignmentCacheKey) -> String {
```
schmit, line 50: Nit: naming is slightly confusing here: the function generates a cache key, but also takes a cache key as argument. 

Maybe it's clearer if we rename `AssignmentCacheKey` to `AssignmentEvent` or `CacheAssignmentEvent`

Sources/eppo/EppoClient.swift
```diff
@@ -142,16 +143,34 @@ public class EppoClient {
...
-            flagKey: flagKey,
-            allocationKey: rule.allocationKey,
-            variation: assignedVariation.value,
-            subject: subjectKey,
-            timestamp: ISO8601DateFormatter().string(from: Date()),
-            subjectAttributes: subjectAttributes
-        )
-        self.assignmentLogger?(assignment);
+        // Optionally log assignment
+        if let assignmentLogger = self.assignmentLogger {
```
schmit, line 147: I know you didn't change this but: why is logging of assignment optional?

schmit (approved):
🔥

Sources/eppo/EppoClient.swift
leoromanovsky: If the logger function is not defined

Sources/eppo/AssignmentCache.swift
leoromanovsky: Good call.

PR #18, opened 2024-04-18
Update to UFC public APIs; bump to version 3.0.0 (FF-1950)

leoromanovsky:
## Context

The upcoming UFC introduces breaking API changes. The work to migrate the ios sdk is being collected in `lr/ff-1938/ufc` - after all sub-tasks are completed (merged into it), it will be merged to the main branch as version `3.0.0`.

Several decisions have been made regarding the public API. The format will generally follow:

```
#  all arguments mandatory: keeps key and attributes together; changes the order compared to today.
getStringAssignment('new-checkout-page', user.id, userAttributes, 'off')
```

The rationale being:

* the flag key is the most obvious identifier for the application code - developers will want to see this first in order to determine the purpose
* keeps subject related arguments closer together
* adds a new default value parameter

## Changes in this PR

* removes `getAssignment` - it was previously deprecated
* updates argument ordering to match decision above
* replaces `argument label omission` style with `required argument label` - I believe this increases clarity when there are consecutive same-type parameters (such as `String, String`). A mistake here has large consequences that we want our users to avoid. One of our unit tests had this 🐛 There are [other libraries](https://docs.launchdarkly.com/sdk/client-side/ios) that also use this pattern.

### DIFF
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -1,6 +1,6 @@
 import Foundation;
 
-public let version = "1.0.1"
+public let version = "3.0.0"
 
 public struct FlagConfigJSON : Decodable {
     var flags: [String : FlagConfig];
@@ -28,7 +28,7 @@ public class EppoClient {
     }
 
     public init(
-        _ apiKey: String,
+        apiKey: String,
         host: String = "https://fscdn.eppo.cloud",
         assignmentLogger: AssignmentLogger? = nil,
         assignmentCache: AssignmentCache? = InMemoryAssignmentCache()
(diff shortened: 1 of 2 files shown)


Sources/eppo/EppoClient.swift
```diff
@@ -52,52 +52,68 @@ public class EppoClient {
...
-        return try getInternalAssignment(subjectKey, flagKey, subjectAttributes, false)?.stringValue()
-    }
 
     public func getBoolAssignment(
-        _ subjectKey: String,
-        _ flagKey: String,
-        _ subjectAttributes: SubjectAttributes = SubjectAttributes()) throws -> Bool?
+        flagKey: String,
+        subjectKey: String,
+        subjectAttributes: SubjectAttributes = SubjectAttributes(),
```
leoromanovsky, line 59: We talked about subject attributes being required - still if anything is going to have a default argument its this. Should we put it to bed and just avoid the default (requiring users to specify the empty object?)

Tests/eppo/ClientTests.swift
```diff
@@ -132,13 +158,13 @@ final class eppoClientTests: XCTestCase {
...
        try super.setUpWithError()
        loggerSpy = AssignmentLoggerSpy()
-       eppoClient = EppoClient("mock-api-key",
+       eppoClient = EppoClient(apiKey: "mock-api-key",
                                host: "http://localhost:4001",
                                assignmentLogger: loggerSpy.logger)
    }
    
    func testUnloadedClient() async throws {
-       XCTAssertThrowsError(try eppoClient.getStringAssignment("badFlagRising", "allocation-experiment-1"))
```
leoromanovsky, line 141: Whoops - the flag and subject arguments are backwards here. labeled parameters and type-enforcement FTW.

Sources/eppo/EppoClient.swift
```diff
@@ -28,7 +28,7 @@ public class EppoClient {
     }
 
     public init(
-        _ apiKey: String,
+        apiKey: String,
```
leoromanovsky, line 31: requiring a label here for consistency

Sources/eppo/EppoClient.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation;
 
-public let version = "1.0.1"
+public let version = "3.0.0"
```
schmit, line 3: 🔥

Sources/eppo/EppoClient.swift
```diff
@@ -52,52 +52,68 @@ public class EppoClient {
...
-    }
 
     public func getBoolAssignment(
-        _ subjectKey: String,
-        _ flagKey: String,
-        _ subjectAttributes: SubjectAttributes = SubjectAttributes()) throws -> Bool?
+        flagKey: String,
+        subjectKey: String,
+        subjectAttributes: SubjectAttributes = SubjectAttributes(),
+        defaultValue: Bool) throws -> Bool?
```
schmit, line 60:
```suggestion
        defaultValue: Bool) throws -> Bool
```

Should always return a bool no?

Sources/eppo/EppoClient.swift
schmit: I think both paths are reasonable -- but maybe for consistency across all SDKs we force it to be specified always. 

In particular, if we do not hten it may be confusing for JSON objects, as subjectAttributes and the default are both dictionaries..., so if only 1 is specified it's really not obvious which argument you are supplying

schmit (approved):

Sources/eppo/EppoClient.swift
leoromanovsky: It's only confusing for dynamically typed languages 😌

Tests/eppo/ClientTests.swift
```diff
@@ -112,13 +138,23 @@ struct AssignmentTestCase : Decodable {
...
-                    $0.subjectKey, self.experiment, $0.subjectAttributes
+                    flagKey: self.experiment,
+                    subjectKey: $0.subjectKey,
+                    subjectAttributes: $0.subjectAttributes,
+                    defaultValue: ""
                 )
             });
         }
 
         if self.subjects != nil {
```
yfrancis, line 149: consider using [optional binding](https://developer.apple.com/documentation/swift/optional#Optional-Binding) with `if let` here:

```
if let subjects = self.subjects {
   return subjects.map({
```

Sources/eppo/EppoClient.swift
```diff
@@ -52,52 +52,68 @@ public class EppoClient {
...
-    {
-        return try getInternalAssignment(subjectKey, flagKey, subjectAttributes, false)?.stringValue()
-    }
 
     public func getBoolAssignment(
-        _ subjectKey: String,
-        _ flagKey: String,
-        _ subjectAttributes: SubjectAttributes = SubjectAttributes()) throws -> Bool?
+        flagKey: String,
+        subjectKey: String,
```
yfrancis, line 58: how important is it to flip the order of subjectKey and flagKey here? Adding the required labels makes this less likely, but I worry that this might be easy to miss in a refactoring

Tests/eppo/ClientTests.swift
leoromanovsky: this is a really nice pattern - swift is a language i wish i used more

PR #19, opened 2024-04-19
Add Sharder (FF-1959)

leoromanovsky:
An aspect of the UFC is computing shards.

Converted the [Python implementation](https://github.com/Eppo-exp/python-sdk/pull/28/files#diff-9b46fdbd999dbff8cb9d10b394020c04cd1e72c73f2d8369b75c4d7fe95193b2) to Swift. Commentary on the implementation is available in-line.

### DIFF
--- a/Sources/eppo/Sharder.swift
+++ b/Sources/eppo/Sharder.swift
@@ -1,0 +1,32 @@
+import Foundation
+import CryptoKit
+
+protocol Sharder {
+    func getShard(input: String, totalShards: Int) -> Int
+}
+
+class MD5Sharder: Sharder {
+    func getShard(input: String, totalShards: Int) -> Int {
+        let inputData = Data(input.utf8)
+        let hash = Insecure.MD5.hash(data: inputData)
+        let hexString = hash.map { String(format: "%02hhx", $0) }.joined()
+
+        // Get the first 8 characters of the MD5 hex string and parse them as an integer using base 16
+        let substringIndex = hexString.index(hexString.startIndex, offsetBy: 8)
+        let intFromHash = Int(hexString[..<substringIndex], radix: 16)!
+
+        return intFromHash % totalShards
+    }
+}
+
+class DeterministicSharder: Sharder {
+    var lookup: [String: Int]
+
+    init(lookup: [String: Int]) {
+        self.lookup = lookup
+    }
+
+    func getShard(input: String, totalShards: Int) -> Int {
+        return lookup[input, default: 0]
+    }
+}
--- a/Tests/eppo/SharderTests.swift
+++ b/Tests/eppo/SharderTests.swift
@@ -6,12 +6,16 @@ class SharderTests: XCTestCase {
     
     func testMD5Sharder() {
         let sharder = MD5Sharder()
-        let input = "testInput"
-        let totalShards = 10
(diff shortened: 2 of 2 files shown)


Sources/eppo/Sharder.swift
```diff
@@ -0,0 +1,32 @@
...
+        let inputData = Data(input.utf8)
+        let hash = Insecure.MD5.hash(data: inputData)
+        let hexString = hash.map { String(format: "%02hhx", $0) }.joined()
+
+        // Get the first 8 characters of the MD5 hex string and parse them as an integer using base 16
+        let substringIndex = hexString.index(hexString.startIndex, offsetBy: 8)
+        let intFromHash = Int(hexString[..<substringIndex], radix: 16)!
+
+        return intFromHash % totalShards
+    }
```
leoromanovsky, line 19: This Swift code is a method within a class that implements the Sharder protocol. The method, getShard, takes two parameters: input, a String, and totalShards, an Int. It returns an Int that represents the shard number calculated based on the input string. Here's a step-by-step breakdown of what the code does:

1. Convert Input to Data: It first converts the input string into a Data object using UTF-8 encoding. This is necessary because the MD5 hashing function operates on bytes rather than on a string directly.

2. Hash the Data: It then computes the MD5 hash of this data. MD5 is a widely used hashing function that produces a 128-bit (16-byte) hash value. It's worth noting that Insecure.MD5 is used here, which Swift marks as insecure because MD5 is not collision-resistant and should not be used for cryptographic purposes.

3. Convert Hash to Hex String: The 16-byte hash is then converted to a hexadecimal string. Each byte is formatted as a two-character hexadecimal number, and these are concatenated to form a single string representing the hash.

4. Extract Integer from Hex String: The code takes the first 8 characters of this hexadecimal string and parses them as an integer using base 16 (hexadecimal). This effectively converts a portion of the hash into an integer. The choice of the first 8 characters is arbitrary and serves to reduce the hash to a manageable integer size while still aiming to maintain some level of distribution.

5. Calculate Shard Number: Finally, it calculates the shard number by taking the modulus of the integer obtained from the hash with the total number of shards 1. The result is the shard number, which is returned by the method.
The purpose of this method is to distribute input strings across a fixed number of shards in a deterministic manner. By using the MD5 hash of the input, it ensures that the same input string always maps to the same shard, while the use of the modulus operation distributes different inputs across the available shards. This technique is commonly used in distributed systems for load balancing or data partitioning.

Tests/eppo/SharderTests.swift
```diff
@@ -0,0 +1,25 @@
...
+    
+    func testMD5Sharder() {
+        let sharder = MD5Sharder()
+        let input = "testInput"
+        let totalShards = 10
+        let shard = sharder.getShard(input: input, totalShards: totalShards)
+        
+        XCTAssertGreaterThanOrEqual(shard, 0)
+        XCTAssertLessThan(shard, totalShards)
+    }
```
schmit, line 15: Added some MD5 sharder tests to JS-commons here, might be nice to copy-paste to ensure they are the same:

https://github.com/Eppo-exp/js-client-sdk-common/pull/52/files#diff-ea83fb75e9f26161475fd7e79c9bd4eb4292dc61b8b02161d43baa036f5d6140

schmit (approved):
🚀

PR #20, opened 2024-04-19
UFC flag evaluator

leoromanovsky:
## motivation

incremental work towards re-launching the UFC.

**scope**

* compatibility with data test

**not in-scope**

* getJSONAssignment - need additional research to understand canonical json handling in swift
* obfuscation - will layer on top


Sources/eppo/ConfigurationRequester.swift
```diff
@@ -2,62 +2,15 @@ import Foundation;
...
     private let httpClient: EppoHttpClient;
 
     public init(httpClient: EppoHttpClient) {
         self.httpClient = httpClient
     }
 
-    public func fetchConfigurations() async throws -> RACConfig {
+    public func fetchConfigurations() async throws -> UniversalFlagConfig {
         let (urlData, _) = try await httpClient.get(RAC_CONFIG_URL);
-        return try ConfigurationRequester.decodeRACConfig(from: String(data: urlData, encoding: .utf8) ?? "");
```
leoromanovsky, line 30: Moved this to the `UniversalFlagConfig` struct as a static method

Sources/eppo/EppoClient.swift
```diff
@@ -97,147 +130,100 @@ public class EppoClient {
...
     }
     
     private func getInternalAssignment(
         flagKey: String,
         subjectKey: String,
         subjectAttributes: SubjectAttributes,
-        useTypedVariationValue: Bool) throws -> EppoValue?
+        expectedVariationType: UFC_VariationType) throws -> FlagEvaluation?
     {
-        try self.validate();
```
leoromanovsky, line 114: functions that aren't re-used don't make sense to me

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
 
 typealias ConditionFunc = (Double, Double) -> Bool;
 
-class Compare {
-    public static func compareNumber(_ a: Double, _ b: Double, _ conditionFunc: ConditionFunc) -> Bool {
```
leoromanovsky, line 7: not used

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
...
+        guard let value = attributeValue else {
+            // Handle the nil case, perhaps throw an error or return a default value
+            return false
+        }
+
         do {
-            switch condition.targetingOperator {
-            case .GreaterThanEqualTo, .GreaterThan, .LessThanEqualTo, .LessThan:
+            switch condition.operator {
+            case .greaterThanEqual, .greaterThan, .lessThanEqual, .lessThan:
```
leoromanovsky, line 199: this casing appears to be canonical enum usage in swift

Sources/eppo/dto/EppoValue.swift
```diff
@@ -2,154 +2,167 @@ import Foundation
...
-
-        return rval;
+    public func isString() -> Bool {
+        return self.type == EppoValueType.String
     }
     
-    public func doubleValue() throws -> Double {
-        if self.value == nil {
-            throw Errors.valueNotSet;
+    public func getBoolValue() throws -> Bool {
```
leoromanovsky, line 121: Leaving this as a `throws` - alternatively can return a `Bool?`. Either way a missing value needs to be handled

Tests/eppo/ClientTests.swift
```diff
@@ -155,43 +103,19 @@ final class eppoClientTests: XCTestCase {
...
-       
-       XCTAssertThrowsError(try eppoClient.getStringAssignment(
-            flagKey: "badFlagRising",
-            subjectKey: "def",
-            subjectAttributes: SubjectAttributes(),
-            defaultValue: ""))
-       {
-           error in XCTAssertEqual(error as! Errors, Errors.flagConfigNotFound)
-       };
-   }
```
leoromanovsky, line 180: handled in centralized tests

Tests/eppo/ConfigurationRequesterTests.swift
```diff
@@ -16,56 +16,6 @@ class ConfigurationRequesterTests: XCTestCase {
         configurationRequester = nil
         super.tearDown()
     }
-
-    func testDecodeRACConfig_ValidRACConfigJSON() throws {
```
leoromanovsky, line 20: this kind of thing is in `UniversalFlagConfigTests`

Sources/eppo/EppoClient.swift
```diff
@@ -55,40 +57,71 @@ public class EppoClient {
...
-        )?.boolValue() ?? defaultValue
+        do {
+            return try getInternalAssignment(
+                flagKey: flagKey, 
+                subjectKey: subjectKey, 
+                subjectAttributes: subjectAttributes,
+                expectedVariationType: UFC_VariationType.boolean
+            )?.variation?.value.getBoolValue() ?? defaultValue
+        } catch {
+            // todo: implement graceful mode
```
aarsilv, line 68: ❤️

Sources/eppo/EppoClient.swift
```diff
@@ -55,40 +57,71 @@ public class EppoClient {
...
+            return defaultValue
+        }
     }
     
-    public func getJSONStringAssignment(
+    // todo: add back when supporting JSON
```
aarsilv, line 73: The UFC does support JSON, transmitted as a JSON string with the variation type JSON. What our SDKs do is attempt to parse and return the native JSON root node. (E.g. no JSON type in EppoValue anymore as its represented as a string and just-in-time parsed)

Sources/eppo/EppoClient.swift
```diff
@@ -97,147 +130,100 @@ public class EppoClient {
...
+            throw Errors.variationTypeMismatch
         }
 
-        guard let assignedVariation = getAssignedVariation(
-            subjectKey, flagKey, flagConfig.subjectShards, allocation.variations
-        ) else {
-            return nil;
+        let flagEvaluation = flagEvaluator.evaluateFlag(flag: flagConfig, subjectKey: subjectKey, subjectAttributes: subjectAttributes)
+        
+        if let variation = flagEvaluation.variation, !isValueOfType(expectedType: expectedVariationType, variationValue: variation.value) {
```
aarsilv, line 174: 👍

Sources/eppo/EppoClient.swift
```diff
@@ -97,147 +130,100 @@ public class EppoClient {
...
+                    // Either the cache is not defined, or the assignment hasn't been logged yet
+                    // Perform assignment.
+                    let assignment = Assignment(
+                        flagKey: flagKey,
+                        allocationKey: allocationKey,
+                        variation: variation.key,
+                        subject: subjectKey,
+                        timestamp: ISO8601DateFormatter().string(from: Date()),
+                        subjectAttributes: subjectAttributes
+                    )
```
aarsilv, line 205: The UFC has two new logging fields, `extra logging` and `metadata,` which we should bake into here as well.

Sources/eppo/RuleEvaluator.swift
aarsilv: :axe:

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
...
+class Compare {    
     public static func compareRegex(_ a: String, _ pattern: String) -> Bool {
         return a.range(of: pattern, options:.regularExpression) != nil;
     }
-
+    
     public static func isOneOf(_ a: String, _ values: [String]) -> Bool {
         return values.map({ $0.lowercased() })
-                     .contains(a.lowercased());
+            .contains(a.lowercased());
```
aarsilv, line 13: UFC is now fully case-sensitive so we should no longer lowercase.

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
...
+        }
+    }
+    
+    private func matchesRules(
+        subjectAttributes: SubjectAttributes,
+        rules: [UFC_Rule]
+    ) -> Bool {
+        if rules.isEmpty {
+            return true
+        }
```
aarsilv, line 146: :+1:

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
...
+
+        // If any rule matches, return true.
```
aarsilv, line 148: :+1:

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
...
+        for allocation in flag.allocations {
+            // # Skip allocations that are not active
+            if let startAt = allocation.startAt, now < startAt {
+                continue
+            }
+            if let endAt = allocation.endAt, now > endAt {
+                continue
+            }
+            
+            if matchesRules(subjectAttributes: subjectAttributes, rules: allocation.rules ?? []) {
```
aarsilv, line 99: An annoying "feature" we need to continue supporting is implicitly passing the subject key as the `id` attribute if an `id` attribute is not specified. Slack thread: https://eppo-group.slack.com/archives/C06N7AA6M5F/p1714767775054549

Sources/eppo/RuleEvaluator.swift
```diff
@@ -3,131 +3,259 @@ import Semver
...
+                } catch let e {
                     // If stringValue() or doubleValue() throws, or Semver creation fails
                     return false
                 }
-            case .Matches:
+            case .matches:
                 return try Compare.compareRegex(
-                    value.stringValue(),
-                    condition.value.stringValue()
+                    value.getStringValue(),
```
aarsilv, line 240: I think you need to change the applied function to `matchesRegex()` as we landed on partial matches by default with regular expressions.

Sources/eppo/dto/EppoValue.swift
aarsilv: Yeah I think missing values should throw

Sources/eppo/dto/EppoValue.swift
```diff
@@ -2,154 +2,167 @@ import Foundation
...
 
     public func toHashedString() -> String {
+        // todo: this is a bit of a hack. we should probably just use a switch
+        // to handle the different types.
         var str = ""
-        if let value = self.value {
+        if let value = self.stringValue {
             str = value
-        } else if let array = self.array {
+        } else if let array = self.stringArrayValue {
```
aarsilv, line 157: I don't know we've aligned on a standard but in Java/Android we have ea space before the comma in our join `" ,"` for string value of array of strings

Tests/eppo/ClientTests.swift
```diff
@@ -7,10 +7,10 @@ import OHHTTPStubsSwift
 @testable import eppo_flagging
 
 let fileURL = Bundle.module.url(
-    forResource: "Resources/test-data/rac-experiments-v3.json",
+    forResource: "Resources/test-data/ufc/flags-v1.json",
```
aarsilv, line 10: 📈

aarsilv: Note this is the unobfuscated version; I imagine a subsequent PR will deal with all the obfuscation fun 😅

Tests/eppo/ClientTests.swift
```diff
@@ -211,25 +135,65 @@ final class eppoClientTests: XCTestCase {
...
-               XCTAssertEqual(assignments, expectedAssignments);
-           case "string":
+               let expectedAssignments = testCase.subjects.map { try? $0.assignment.getBoolValue() }
+               
+               if assignments != expectedAssignments {
+                    let difference = zip(assignments, expectedAssignments)
+                        .filter { $0 != $1 }
+                        .map { "(\($0), \($1))" }
+                        .joined(separator: ", ")
+                    XCTFail("Mismatch for flagKey: \(testCase.flag) - Differences: [\(difference)]")
```
aarsilv, line 148: ❤️ I appreciate the helpful failure message (and future devs will too!)

Tests/eppo/ClientTests.swift
```diff
@@ -211,25 +135,65 @@ final class eppoClientTests: XCTestCase {
...
+                } else {
+                    XCTAssert(true)
+                }
+           case UFC_VariationType.json:
+                print("json not supported")
```
aarsilv, line 153: AU CONTRAIRE MON FRÈRE

Tests/eppo/ConfigurationRequesterTests.swift
aarsilv: Love that more coverage is moving towards our shared test suite!

Tests/eppo/RuleEvaluatorTests.swift
```diff
@@ -2,195 +2,257 @@ import XCTest
...
+final class flagEvaluationTests: XCTestCase {
+    public func testNoneResult() {
+        let flagEvaluation = FlagEvaluation.noneResult(
+            flagKey: "test",
+            subjectKey: "test",
+            subjectAttributes: SubjectAttributes()
+        );
+        XCTAssertEqual(flagEvaluation.flagKey, "test");
+        XCTAssertEqual(flagEvaluation.subjectKey, "test");
+        //XCTAssertEqual(flagEvaluation.subjectAttributes, SubjectAttributes());
```
aarsilv, line 14: why is this commented out?

Tests/eppo/RuleEvaluatorTests.swift
```diff
@@ -2,195 +2,257 @@ import XCTest
...
+        let evaluationResult = flagEvaluator.evaluateFlag(flag: testFlag, subjectKey: "subject_key", subjectAttributes: subjectAttributes)
+        XCTAssertTrue(evaluationResult.doLog)
     }
     
-    public func testMatchesAnyRuleWhenThrowInvalidSubjectAttribute() {
-        var targetingRule: TargetingRule = self.createRule([]);
-        targetingRule.conditions.append(contentsOf: self.getNumericConditions());
-        let targetingRules: [TargetingRule] = [targetingRule];
-        
+    // TODO: what is the behavior here? I think it should not match numeric
```
aarsilv, line 96: If it doesn't match I agree `doLog` should be `false`

Tests/eppo/RuleEvaluatorTests.swift
```diff
@@ -2,195 +2,257 @@ import XCTest
...
-        let targetingRules: [TargetingRule] = [];
+        let targetingRules: [UFC_Rule] = [];
+        let testFlag = createFlag(flag: flag, rules: targetingRules);
+
         var subjectAttributes: SubjectAttributes = SubjectAttributes();
         subjectAttributes["name"] = EppoValue.valueOf("test");
         
-        XCTAssertNil(try RuleEvaluator.findMatchingRule(subjectAttributes, targetingRules));
+        let evaluationResult = flagEvaluator.evaluateFlag(flag: testFlag, subjectKey: "subject_key", subjectAttributes: subjectAttributes)
+        XCTAssertTrue(evaluationResult.doLog)
```
aarsilv, line 69: While we don't need to assert every field of the evaluation result since we cover that with the first test, I think for all these other tests we'll still want to assert `variation` in addition to `do Log`.

Tests/eppo/RuleEvaluatorTests.swift
```diff
@@ -2,195 +2,257 @@ import XCTest
...
+        XCTAssertNil(flagEvaluation.variation);
+        XCTAssertEqual(flagEvaluation.extraLogging, [:]);
+        XCTAssertFalse(flagEvaluation.doLog);
+    }
+}
+
+final class flagEvaluatorTests: XCTestCase {
+    var flagEvaluator: FlagEvaluator!
+
+    let flag = UFC_Flag(
```
aarsilv, line 25: maybe name this `baseFlag` to indicate its a starting point for tests

Tests/eppo/RuleEvaluatorTests.swift
```diff
@@ -2,195 +2,257 @@ import XCTest
...
-        return condition;
+    private func getNotOneOfCondition() -> UFC_TargetingRuleCondition {
+        return UFC_TargetingRuleCondition(
+            operator: UFC_RuleConditionOperator.notOneOf,
+            attribute: "oneOf", // one of or not one off??
+            value: EppoValue.valueOf(["value1", "value2"])
+        );
     }
+
+    private func createFlag(flag: UFC_Flag, rules: [UFC_Rule], enabled: Bool = true) -> UFC_Flag {
```
aarsilv, line 220: same here I'd name the first arg `baseFlag`

Sources/eppo/RuleEvaluator.swift
schmit: Adding a test for this in the sdk-test-data repo

Sources/eppo/RuleEvaluator.swift
schmit: Adding a test for this in the sdk-test-data repo

Sources/eppo/EppoClient.swift
leoromanovsky: Yep understood - mostly I wanted to get the logic in this PR up for review while I figured out what to do about json serialization in swift.

Sources/eppo/RuleEvaluator.swift
leoromanovsky: @schmit would modifying the test files be possible here as well?

Sources/eppo/RuleEvaluator.swift
leoromanovsky: I see this in python too: https://github.com/Eppo-exp/python-sdk/blob/main/eppo_client/rules.py#L59-L61

Sources/eppo/RuleEvaluator.swift
leoromanovsky: Do you have an example of this? Checked python and android without seeing it.

Sources/eppo/RuleEvaluator.swift
leoromanovsky: the current implementation:

```
return a.range(of: pattern, options:.regularExpression) != nil;
```

already supports partial matching. I will rename the wrapper method for clarity.

leoromanovsky:
critical changes from feedback:

* case-sensitive comparison
* validated that regex matching is happening

other changes:

* metadata and extra logging into assignment
* tidy up

Tests/eppo/ClientTests.swift
```diff
@@ -144,97 +43,114 @@ final class eppoClientTests: XCTestCase {
...
-       }
-       
-       XCTAssertGreaterThan(testFiles.count, 0);
-   }
+            defaultValue: 0)
+        XCTAssertEqual(assignment, 3.1415926)
+        XCTAssertTrue(loggerSpy.wasCalled)
+        if let lastAssignment = loggerSpy.lastAssignment {
+            XCTAssertEqual(lastAssignment.allocation, "rollout")
+            XCTAssertEqual(lastAssignment.experiment, "numeric_flag-rollout")
```
leoromanovsky, line 69: @yfrancis We have a test for the android failure that was observed.

Sources/eppo/EppoClient.swift
```diff
@@ -97,147 +130,102 @@ public class EppoClient {
...
+                    // The assignment has already been logged, do nothing
+                } else {
+                    // Either the cache is not defined, or the assignment hasn't been logged yet
+                    // Perform assignment.
+                    let assignment = Assignment(
+                        flagKey: flagKey,
+                        allocationKey: allocationKey,
+                        variation: variation.key,
+                        subject: subjectKey,
+                        timestamp: ISO8601DateFormatter().string(from: Date()),
```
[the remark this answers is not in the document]
leoromanovsky: Are you suggesting getting the current date at the top of the function or something else?

leoromanovsky:
New broken test cases were introduced as this code was being developers. They will be fixed by upstream PRs:

* https://github.com/Eppo-exp/eppo-ios-sdk/pull/29
* https://github.com/Eppo-exp/eppo-ios-sdk/pull/27

Sources/eppo/EppoClient.swift
```diff
@@ -55,40 +57,71 @@ public class EppoClient {
...
+//    }
+//    
+    
+    public func getIntegerAssignment(
         flagKey: String,
         subjectKey: String,
-        subjectAttributes: SubjectAttributes,
-        defaultValue: String) throws -> String
+        subjectAttributes: SubjectAttributes = SubjectAttributes(),
+        defaultValue: Int) throws -> Int
```
schmit, line 92: No expert on Swift, so just double checking -- this function should not throw errors right?

leoromanovsky: It does throw so the caller has to handle that; the code won't compile for them with a handle.

We added the safe mode to other SDKs because those don't have the throwable trait; Java has this as well. So developers won't be surprised by unexpected exceptions.

leoromanovsky: Plenty of them - just to name a few if a `flagKey` or `subjectKey` is empty and others. I haven't yet added explicit support for graceful mode in swift but I am thinking it is not as necessary as in other SDKs since the problem graceful mode was trying to solve was when users were surprised by an exception. By having a `throws -> Int` type they will now to handle the return from assignment methods.

We can't have it both ways though: if we wrap each public method in a `do/catch` block and want to remove `throws` then the `catch` has to mask all exceptions. Conditionally throwing is still throwing.

Is there something I am missing or a concern that exceptions will leak unexpectedly to users?

yfrancis: I ask because the body of the method is wrapped in a do/catch, so my expectation would be that all exceptions are handled within, and that callers of this method have no exceptions to handle themselves

leoromanovsky: That's fair - what I meant was I want those exceptions to be returned and have a follow up (https://linear.app/eppo/issue/FF-1953/add-graceful-mode-to-ios-sdk) to address it. 

We'll need to allow exceptions to be thrown if we want to use that as a signaling method when something went wrong, or opt for log messages and always return the default message (current behavior).

yfrancis: Gotcha, that plan makes sense to me, thanks for clarifying!

Sources/eppo/dto/EppoValue.swift
```diff
@@ -2,154 +2,171 @@ import Foundation
...
+        // todo: this is a bit of a hack. we should probably just use a switch
+        // to handle the different types.
         var str = ""
-        if let value = self.value {
+        if let value = self.stringValue {
             str = value
-        } else if let array = self.array {
-            str = array.joined(separator: ",")
+        } else if let array = self.stringArrayValue {
+            str = array.joined(separator: " ,")
```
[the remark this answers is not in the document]
aarsilv: Shoot yeah my comment was bad, this is what we want 😅

Tests/eppo/RuleEvaluatorTests.swift
```diff
@@ -2,195 +2,266 @@ import XCTest
...
+        let targetingRules: [UFC_Rule] = [];
+        let testFlag = createFlag(flag: baseFlag, rules: targetingRules);
+
         var subjectAttributes: SubjectAttributes = SubjectAttributes();
         subjectAttributes["name"] = EppoValue.valueOf("test");
         
-        XCTAssertNil(try RuleEvaluator.findMatchingRule(subjectAttributes, targetingRules));
+        let evaluationResult = flagEvaluator.evaluateFlag(flag: testFlag, subjectKey: "subject_key", subjectAttributes: subjectAttributes)
+        XCTAssertTrue(evaluationResult.doLog)
+        XCTAssertEqual(try evaluationResult.variation?.value.getStringValue(), "Control")
```
aarsilv, line 72: ✅

aarsilv (approved):
Thanks for iterating! Things are shaping up with this iterative approach. Approving this iteration, although be sure to put in the comma order change for comparing lists of strings that Sven caught!

Sources/eppo/EppoClient.swift
```diff
@@ -97,147 +130,102 @@ public class EppoClient {
...
+            if let allocationKey = flagEvaluation.allocationKey,
+               let variation = flagEvaluation.variation,
+               let assignmentLogger = self.assignmentLogger {
+                
+                // Prepare the assignment cache key
+                let assignmentCacheKey = AssignmentCacheKey(
+                    subjectKey: subjectKey,
+                    flagKey: flagKey,
+                    allocationKey: allocationKey,
+                    variationValue: variation.value
```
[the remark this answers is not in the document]
leoromanovsky: I will look into that separately, thank you

PR #21, opened 2024-04-25
v3.0.0 UFC

leoromanovsky:
## Motivation

Support for the UFC migration.

## Description

* Supports all UFC operations, validated by shared test files.
* Obfuscation support.

All of these changes have previously been reviewed.


Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,159 @@
...
+        switch self {
+        case .notUTF8Encoded(let message):
+            return message
+        case .parsingError(let message):
+            return message
+        }
+    }
+}
+
+enum UFC_VariationType: String, Decodable {
```
felipecsl, line 73: curious about your naming decision on the `UFC_` prefix, haven't seen that in other SDKs. could we keep camelCase for consistently?

felipecsl (approved):
Cursory look seems fine, just one question on naming
Also are we completely dropping RAC? What if we need to make a bug fix for it in the future, I assume we can patch the v2.x separately if needed, for users still on that version?

Sources/eppo/UniversalFlagConfig.swift
leoromanovsky: I did this because I had both the old models and the new UFC models co-existing and there was some conflict in the names, such as `Variant`

PR #25, opened 2024-06-01
Parse UFC API into native models (FF-2311)

leoromanovsky:
## motivation

* SDK is being migrated to the UFC API - my goal is to keep the PRs small so that reviews are quick and engineers can keep their attention fresh.
* 🐛 `value` field can handle incoming integer values to `EppoValue` but not `Double`

```
{
   value: 123.45
}
```

* Base branch continues to be `lr/ff-1938/ufc` not `main`. After all PRs are accepted and tests pass we will merge the entire feature into main.

## changes

* Adds new models for parsing UFC API; these are in parallel with the existing RAC models and have no impact on the production code.
* Fixes bug to allow parsing `Double` to the `EppoValue` class.
* Adds new unit test for `EppoValue`.
* More granular errors returned during json parsing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,5 @@ test-data:
 	rm -rf $(testDataDir)
 	mkdir -p $(tempDir)
 	git clone -b ${branchName} --depth 1 --single-branch ${githubRepoLink} ${gitDataDir}
-	cp ${gitDataDir}rac-experiments-v3.json ${testDataDir}
-	cp -r ${gitDataDir}assignment-v2 ${testDataDir}
+	cp -r ${gitDataDir} ${testDataDir}
 	rm -rf ${tempDir}
--- a/Sources/eppo/UniversalFlagConfig.swift
+++ b/Sources/eppo/UniversalFlagConfig.swift
@@ -81,6 +81,17 @@ enum UFC_AlgorithmType: String, Decodable {
     case contextualBandit = "CONTEXTUAL_BANDIT"
 }
 
+enum UFC_RuleConditionOperator: String, Decodable {
+  case lessThan = "LT"
+  case lessThanEqual = "LTE"
+  case greaterThan = "GT"
+  case greaterThanEqual = "GTE"
+  case matches = "MATCHES"
+  case oneOf = "ONE_OF"
+  case notOneOf = "NOT_ONE_OF"
+  case isNull = "IS_NULL"
+}
+
 // models
 
 public struct UFC_Flag : Decodable {
@@ -112,7 +123,7 @@ public struct UFC_Rule : Decodable {
 }
 
 public struct UFC_TargetingRuleCondition : Decodable {
-    let `operator`: String;
+    let `operator`: UFC_RuleConditionOperator;
     let attribute: String;
     let value: EppoValue;
 }
--- a/Sources/eppo/UniversalFlagConfig.swift
+++ b/Sources/eppo/UniversalFlagConfig.swift
@@ -1,0 +1,134 @@
+import Foundation
+
+public struct UniversalFlagConfig : Decodable {
+    let createdAt: Date?;
+    let flags: [String: UFC_Flag];
+    // todo: add bandits
+    
+    static func decodeFromJSON(from jsonString: String) throws -> UniversalFlagConfig {
+        guard let jsonData = jsonString.data(using: .utf8) else {
+            throw UniversalFlagConfigError.notUTF8Encoded("Failed to encode JSON string into UTF-8 data.")
+        }
+
+        let decoder = JSONDecoder()
+        
+        // Set up the date formatter
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"  // Adjusted to include milliseconds
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")  // Use POSIX to avoid unexpected behaviors
+        dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)  // Adjust if your JSON dates are not in GMT
+        
+        // Use the date formatter in the decoder
+        decoder.dateDecodingStrategy = .formatted(dateFormatter)
+        
+        do {
+            return try decoder.decode(UniversalFlagConfig.self, from: jsonData)
+        } catch let error as DecodingError {
+            switch error {
+            case .typeMismatch(_, let context):
+                throw UniversalFlagConfigError.parsingError("Type mismatch: \(context.debugDescription)")
+            case .valueNotFound(_, let context):
+                throw UniversalFlagConfigError.parsingError("Value not found: \(context.debugDescription)")
+            case .keyNotFound(let key, let context):
+                throw UniversalFlagConfigError.parsingError("Key not found: \(key.stringValue) - \(context.debugDescription)")
+            case .dataCorrupted(let context):
+                throw UniversalFlagConfigError.parsingError("Data corrupted: \(context.debugDescription)")
+            default:
+                throw UniversalFlagConfigError.parsingError("JSON parsing error: \(error.localizedDescription)")
+            }
+        } catch {
+            throw UniversalFlagConfigError.parsingError("Unexpected error: \(error.localizedDescription)")
+        }
+    }
+}
+
+// enums
+enum UniversalFlagConfigError: Error, CustomNSError, LocalizedError {
+    case notUTF8Encoded(String)
+    case parsingError(String)
+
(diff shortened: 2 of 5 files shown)


Sources/eppo/dto/EppoValue.swift
```diff
@@ -10,9 +10,9 @@ public enum EppoValueType {
 }
 
 public class EppoValue : Decodable, Equatable {
-    public var value: String?;
-    public var type: EppoValueType = EppoValueType.Null;
-    public var array: [String]?;
+    private var value: String?;
+    private var type: EppoValueType = EppoValueType.Null;
+    private var array: [String]?;
```
leoromanovsky, line 15: 🤫

Sources/eppo/dto/EppoValue.swift
```diff
@@ -54,32 +54,28 @@ public class EppoValue : Decodable, Equatable {
...
-        if self.value != nil {
-            self.type = EppoValueType.Number;
-            return;
-        }
 
-        try? self.value = String(container.decode(Bool.self));
-        if self.value != nil {
-            self.type = EppoValueType.Boolean;
-            return;
+        if let array = try? container.decode([String].self) {
```
leoromanovsky, line 58: collapsed previous code into these one-liner conditionals - swift is great! 😃

Sources/eppo/dto/EppoValue.swift
```diff
@@ -54,32 +54,28 @@ public class EppoValue : Decodable, Equatable {
...
+            self.type = .ArrayOfStrings
+            self.array = array
+            self.value = nil
+        } else if let numericValue = try? container.decode(Double.self) {
+            // decode double handles both integers and floating-point numbers.
```
leoromanovsky, line 63: there is a unit test to demonstrate parsing integers and doubles

Tests/eppo/UniversalFlagConfigTests.swift
```diff
@@ -0,0 +1,47 @@
...
+            withExtension: ""
+        )
+        do {
+            UFCTestJSON = try String(contentsOfFile: fileURL.path)
+        } catch {
+            XCTFail("Error loading test JSON: \(error)")
+        }
+    }
+
+    func testDecodeUFCConfig() {
```
leoromanovsky, line 24: I'm not trying to be exhaustive here just touch some of the models to make sure they are decoding as expected. We need to count on the json parsed throwing an error to truly detect mismatches.

Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,84 @@
...
+
+public struct UFC_Variation : Decodable  {
+    let key: String;
+    let value: EppoValue;
+    let algorithmType: String? // todo: enum
+}
+
+public struct UFC_Allocation : Decodable {
+    let key: String;
+    let rules: [UFC_Rule]?;
```
leoromanovsky, line 53: I kind of wish this was required - seems like it invites room for error. Should be fine here; some of the test files have this missing and unit tests would catch it.

Makefile
```diff
@@ -43,6 +43,5 @@ test-data:
 	rm -rf $(testDataDir)
 	mkdir -p $(tempDir)
 	git clone -b ${branchName} --depth 1 --single-branch ${githubRepoLink} ${gitDataDir}
-	cp ${gitDataDir}rac-experiments-v3.json ${testDataDir}
-	cp -r ${gitDataDir}assignment-v2 ${testDataDir}
```
leoromanovsky, line 47: this was only pulling RAC-based test files. generalize to load all.

Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,84 @@
...
+    let createdAt: Date?;
+    let flags: [String: UFC_Flag];
+    // todo: add bandits
+    
+    static func decodeFromJSON(from jsonString: String) throws -> UniversalFlagConfig {
+        let decoder = JSONDecoder()
+
+        // Set up the date formatter
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"  // Adjusted to include milliseconds
```
leoromanovsky, line 13: the UFC API returns dates with milliseconds:

```
new Date().toISOString()
```

Screenshot 2024-05-31 at 11 48 07 PM

Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+
+public struct UniversalFlagConfig : Decodable {
+    let createdAt: Date?;
+    let flags: [String: UFC_Flag];
```
leoromanovsky, line 5: I am prefixing all models with `UFC_` to avoid collisions with RAC-models. We can drop the at the end of the project but I personally don't mind it.

Makefile
schmit: I switched it to only loading ufc files since the rac tests are not relevant in other SDKs

Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,145 @@
...
+      case integer = "INTEGER"
+      case json = "JSON"
+      case numeric = "NUMERIC"
+      case string = "STRING"    
+}
+
+enum UFC_AlgorithmType: String, Decodable {
+    case constant = "CONSTANT"
+    case contextualBandit = "CONTEXTUAL_BANDIT"
+}
```
schmit, line 82: This is not part of the UFC so we should not need this type

Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,145 @@
...
+    let variationType: UFC_VariationType;
+    let variations: [String: UFC_Variation];
+    let allocations: [UFC_Allocation];
+    let totalShards: Int;
+}
+
+public struct UFC_Variation : Decodable  {
+    let key: String;
+    let value: EppoValue;
+    let algorithmType: UFC_AlgorithmType?
```
schmit, line 109: Can remove this

Sources/eppo/UniversalFlagConfig.swift
```diff
@@ -0,0 +1,145 @@
...
+}
+
+public struct UFC_Rule : Decodable {
+    let conditions: [UFC_TargetingRuleCondition];
+}
+
+public struct UFC_TargetingRuleCondition : Decodable {
+    let `operator`: UFC_RuleConditionOperator;
+    let attribute: String;
+    let value: EppoValue;
```
schmit, line 128: What is EppoValue for you? In other SDKs it refers to the values that the SDK can return and does not include Array[String]

Tests/eppo/UniversalFlagConfigTests.swift
schmit: Makes sense, ideally we can count on the general test cases to provide good testing coverage

schmit (approved):
Appreciate the bite-sized PRs!

Makefile
leoromanovsky: Yep but this PR is additive so I need to continue testing RAC files here.

Sources/eppo/UniversalFlagConfig.swift
leoromanovsky: I got it from the UFC API interface definition. Are you saying that interface is wrong or this is out of scope of the UFC. If the latter yes I know but it doesn't hurt to parse it.

Sources/eppo/UniversalFlagConfig.swift
leoromanovsky: It handles all types: numeric, string and array of strings; originally created for the variation value and typed values. 

I noticed in the python implementation that this was split into two classes to more directly match the API UFC response so they were more limited but as soon as you are returning multiple types in a json value it's not possible to know what to expect so I am accepting all.

Tests/eppo/UniversalFlagConfigTests.swift
leoromanovsky: Yes I believe the test files exercise all permutations in one way or another: some have no rules, variety of splits, etc. I implemented off the UFC API spec instead of the test files so that should cover all cases. Going to gRPC will be even better 🤓

Sources/eppo/UniversalFlagConfig.swift
schmit: It's not part of the config, see https://github.com/Eppo-exp/sdk-test-data/blob/main/ufc/flags-v1.json for the canonical UFC config definition

PR #27, opened 2024-06-04
Add support for getJSONAssignment (FF-1951)

leoromanovsky:
## motivation

Add support for the `json` variation type. Given we cannot know what json parsing library our customers wish to use, we have decided its safer to return the stringified value directly and let them handle as needed.

## description

Add  `getJSONAssignment` that returns a `String`

### DIFF
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -70,7 +70,7 @@ public class EppoClient {
         }
     }
     
-    public func getJSONAssignment(
+    public func getJSONStringAssignment(
         flagKey: String,
         subjectKey: String,
         subjectAttributes: SubjectAttributes,
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -105,7 +105,7 @@ public class EppoClient {
         }
     }
     
-    public func getDoubleAssignment(
+    public func getNumericAssignment(
         flagKey: String,
         subjectKey: String,
         subjectAttributes: SubjectAttributes = SubjectAttributes(),
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -105,7 +105,7 @@ public class EppoClient {
         }
     }
     
-    public func getNumericAssignment(
+    public func getDoubleAssignment(
         flagKey: String,
         subjectKey: String,
         subjectAttributes: SubjectAttributes = SubjectAttributes(),
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -70,20 +70,20 @@ public class EppoClient {
         }
     }
     
-    // todo: add back when supporting JSON
-//    public func getJSONAssignment(
-//        flagKey: String,
-//        subjectKey: String,
-//        subjectAttributes: SubjectAttributes,
-//        defaultValue: [String: EppoValue]) throws -> [String: EppoValue]
-//    {
-//        return try getInternalAssignment(
-//            flagKey: flagKey, 
-//            subjectKey: subjectKey, 
-//            subjectAttributes: subjectAttributes
-//        )?.variation?.value?.objectValue() ?? defaultValue
-//    }
-//    
+    public func getJSONAssignment(
+        flagKey: String,
+        subjectKey: String,
+        subjectAttributes: SubjectAttributes,
+        defaultValue: String) throws -> String
+    {
+        return try getInternalAssignment(
+            flagKey: flagKey, 
+            subjectKey: subjectKey, 
+            subjectAttributes: subjectAttributes,
+            expectedVariationType: UFC_VariationType.json
+        )?.variation?.value.getStringValue() ?? defaultValue
+    }
+    
     
     public func getIntegerAssignment(
         flagKey: String,
--- a/Sources/eppo/RuleEvaluator.swift
+++ b/Sources/eppo/RuleEvaluator.swift
@@ -196,7 +196,13 @@ public class FlagEvaluator {
             // Handle the nil case, perhaps throw an error or return a default value
             return false
         }
-        
+
+        // Safely unwrap attributeValue for further use
+        guard let value = attributeValue else {
+            // Handle the nil case, perhaps throw an error or return a default value
+            return false
+        }
+
         do {
             switch condition.operator {
             case .greaterThanEqual, .greaterThan, .lessThanEqual, .lessThan:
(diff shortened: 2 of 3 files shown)


Sources/eppo/EppoClient.swift
```diff
@@ -70,20 +70,20 @@ public class EppoClient {
...
-//        defaultValue: [String: EppoValue]) throws -> [String: EppoValue]
-//    {
-//        return try getInternalAssignment(
-//            flagKey: flagKey, 
-//            subjectKey: subjectKey, 
-//            subjectAttributes: subjectAttributes
-//        )?.variation?.value?.objectValue() ?? defaultValue
-//    }
-//    
+    public func getJSONAssignment(
```
aarsilv, line 73: I think this is what we want our `getJSONStringAssignment()` to be, but will have a `getJSONAssignment()` that still returns a native JSON assignment, as Swift--and most languages--have native JSON support.

aarsilv:
> Add support for the json variation type. Given we cannot know what json parsing library our customers wish to use, we have decided its safer to return the stringified value directly and let them handle as needed.

I don't think this is where we landed. Many languages have built-in JSON support in our effort to be as "friction free" for developers as possible, we should have a `getJSONAssignment()` that returns it in the natively-parsed way. For example, in our JavaScript SDKs it will just return the JS Object.

We _could_ stop there, as these all have `toString()`-type functionality, but I think we bring back `getJSONStringAssignment()` in case folks do want to do their own parsing for whatever reason.

Sources/eppo/EppoClient.swift
leoromanovsky: Swift would be a [String: any] - this is a dictionary with string keys and unknown values. What else can the shape be without a struct to deserialize to? Do you think this offers a valuable object that users can parse? Easy enough to return this though.

schmit:
I agree with Aaron that it makes more sense for consistency sake to have a separate `getJSONStringAssignment` that returns the JSON as a string for further parsing.

leoromanovsky:
> I agree with Aaron that it makes more sense for consistency sake to have a separate `getJSONStringAssignment` that returns the JSON as a string for further parsing.

Sounds good - I gave a try to returning `[String: Any]` for a few hours but still blocked on something that is kind of frustrating because it isn't related to the real API spec. Our API spec has stringified JSON, which is easier to handle but the test file has native JSON objects:

```
{
    "flag": "json-config-flag",
    "variationType": "JSON",
    "defaultValue": {},
    "subjects": [
      {
        "subjectKey": "alice",
        "subjectAttributes": {"email": "alice@mycompany.com", "country": "US"},
        "assignment": { "integer": 1, "string": "one", "float": 1.0 }
      },
      {
        "subjectKey": "bob",
        "subjectAttributes": {"email": "bob@example.com", "country": "Canada"},
        "assignment": { "integer": 2, "string": "two", "float": 2.0 }
      },
      {
        "subjectKey": "charlie",
        "subjectAttributes": {"age": 50},
        "assignment": { "integer": 2, "string": "two", "float": 2.0 }
      }
    ]
  }
```

So to get a native object returned we also need to add an arbitrary parsed. I would like to move on from this as I don't think the value is very high to return a native type compared to shipping the UFC, and will leave it as a follow up.

leoromanovsky:
Test is failing because of: https://github.com/Eppo-exp/sdk-test-data/pull/25 

I will fix separetely.

Tests/eppo/UniversalFlagConfigTests.swift
```diff
@@ -22,9 +22,6 @@ final class UniversalFlagConfigTest: XCTestCase {
     }
 
     func testDecodeUFCConfig() {
-        let config = try! UniversalFlagConfig.decodeFromJSON(from: UFCTestJSON)
-        XCTAssertEqual(config.flags.count, 14, "There should be 14 flags present in the JSON test file.")
```
leoromanovsky, line 26: not a good assertion, it always changes

schmit: 😬 yup no reason to test for this either; if there is an issue parsing the flags some other tests will fail

schmit (approved):
👍 , we can figure out whether/how to return the parsed JSON object when needed

Sources/eppo/EppoClient.swift
```diff
@@ -70,20 +70,20 @@ public class EppoClient {
...
-//        defaultValue: [String: EppoValue]) throws -> [String: EppoValue]
-//    {
-//        return try getInternalAssignment(
-//            flagKey: flagKey, 
-//            subjectKey: subjectKey, 
-//            subjectAttributes: subjectAttributes
-//        )?.variation?.value?.objectValue() ?? defaultValue
-//    }
-//    
+    public func getJSONStringAssignment(
```
schmit, line 73: 🙏

aarsilv:
> I gave a try to returning `[String: Any]` 

Yes I think this is what we want for swift

> the test file has native JSON objects:

This got me as well in Android land 😞 I ended up needing to make a [`TestCaseValue`](https://github.com/Eppo-exp/android-sdk/blob/main/eppo/src/androidTest/java/cloud/eppo/android/helpers/TestCaseValue.java) which was an extension of `EppoValue` to handle JSON, and then that was used in an [`AssignmentTestCase` class and it's `AssignmentTestCaseAdapter` (deserializer)](https://github.com/Eppo-exp/android-sdk/blob/main/eppo/src/androidTest/java/cloud/eppo/android/helpers/AssignmentTestCaseDeserializer.java#L19).

I wonder if we should have the test cases have the expected assignment as a JSON string as well, and leave it to the individual SDK tests to do the JSON (order of properties doesn't matter) comparison.

aarsilv (approved):
Approving as-is with the `getJSONStringAssignment()` API interface, but would love a `getJSONAssignment()` that returns `[String: Any]` in the future.

leoromanovsky:
> > I gave a try to returning `[String: Any]`
> 
> Yes I think this is what we want for swift
> 
> > the test file has native JSON objects:
> 
> This got me as well in Android land 😞 I ended up needing to make a [`TestCaseValue`](https://github.com/Eppo-exp/android-sdk/blob/main/eppo/src/androidTest/java/cloud/eppo/android/helpers/TestCaseValue.java) which was an extension of `EppoValue` to handle JSON, and then that was used in an [`AssignmentTestCase` class and it's `AssignmentTestCaseAdapter` (deserializer)](https://github.com/Eppo-exp/android-sdk/blob/main/eppo/src/androidTest/java/cloud/eppo/android/helpers/AssignmentTestCaseDeserializer.java#L19).
> 
> I wonder if we should have the test cases have the expected assignment as a JSON string as well, and leave it to the individual SDK tests to do the JSON (order of properties doesn't matter) comparison.

That's a good suggestion - I would save the SDKs from having to do work (parsing native json) that they have no need to. Obviously this is easier in javascript with `JSON.loads`

PR #29, opened 2024-06-07
Adds toEppoString representation of EppoValue; handles (FF-2345)

leoromanovsky:
New test was introduced in: https://github.com/Eppo-exp/sdk-test-data/pull/25

To perform comparison between different types we need a consistent representation across SDKs.

## description

We decided on a common protocol for comparing stringified numbers, integers and decimals. This PR implements that behavior in `toEppoString`

## questions for reviews

some TODOs in-line - given the interface of swift I need to handle `throw` or allow the exception to re-bubble up.

### DIFF
--- a/Sources/eppo/dto/EppoValue.swift
+++ b/Sources/eppo/dto/EppoValue.swift
@@ -14,14 +14,15 @@ public class EppoValue : Decodable, Equatable {
     private var boolValue: Bool?;
     private var doubleValue: Double?
     private var stringValue: String?
-    private var stringArrayValue: [String]?;
+    private var stringArrayValue: [String]?
+    private var objectValue: [String: JSONValue]?
 
     enum Errors : Error {
         case valueNotSet;
     }
 
     public static func == (lhs: EppoValue, rhs: EppoValue) -> Bool {
-        if lhs.type != rhs.type { return false; }
+        if lhs.type != rhs.type { return false }
 
         switch lhs.type {
             case .Boolean:
@@ -66,13 +67,12 @@ public class EppoValue : Decodable, Equatable {
     }
 
     required public init(from decoder: Decoder) throws {
-        let container = try decoder.singleValueContainer();
+        let container = try decoder.singleValueContainer()
 
         if let array = try? container.decode([String].self) {
             self.type = .ArrayOfStrings
             self.stringArrayValue = array
         } else if let doubleValue = try? container.decode(Double.self) {
-            // decode double handles both integers and floating-point numbers.
             self.type = .Numeric
             self.doubleValue = doubleValue
         } else if let boolValue = try? container.decode(Bool.self) {
@@ -83,6 +83,11 @@ public class EppoValue : Decodable, Equatable {
             self.stringValue = stringValue
         } else {
             self.type = .Null
+            self.boolValue = nil
+            self.doubleValue = nil
+            self.stringValue = nil
+            self.stringArrayValue = nil
+            self.objectValue = nil
         }
     }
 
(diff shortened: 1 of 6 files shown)


Tests/eppo/EppoValueTests.swift
```diff
@@ -68,6 +68,23 @@ class EppoValueTests: XCTestCase {
         XCTAssertFalse(arrayValue1 == arrayValueDifferent, "Arrays with different elements or counts should not be equal")
     }
 
+    func testToString() {
```
leoromanovsky, line 71: same tests as in: https://github.com/Eppo-exp/python-sdk/pull/51

missing anything?

Sources/eppo/dto/EppoValue.swift
```diff
@@ -152,21 +152,52 @@ public class EppoValue : Decodable, Equatable {
...
+            } else {
+                // todo: should this throw?
+                return "null"
+            }
+        case .Null:
+            return "null"
         }
+    }
 
+    public func toHashedString() -> String {
```
leoromanovsky, line 194: this function is used for the assignment logging to keep a consistent cache entry size (https://github.com/Eppo-exp/eppo-ios-sdk/blob/main/Sources/eppo/AssignmentCache.swift#L30). it's probably extraneous and I could just use the new function in this PR. Thoughts welcome.

Sources/eppo/dto/EppoValue.swift
```diff
@@ -152,21 +152,52 @@ public class EppoValue : Decodable, Equatable {
...
+                    // If the double value is an integer, return as integer string
+                    return String(format: "%.0f", doubleVal)
+                } else {
+                    // Otherwise, return the double as is
+                    return String(doubleVal)
+                }
+            } else {
+                // todo: should this throw?
+                return "null"
+            }
```
schmit, line 171: When would the self.getDoubleValue fail?

Sources/eppo/dto/EppoValue.swift
```diff
@@ -152,21 +152,52 @@ public class EppoValue : Decodable, Equatable {
...
+            // todo: should this throw?
+            // it should not since we checked the type is a string above
+            // but to be true to the type system we should throw here
+            if let value = try? self.getStringValue() {
+                return value
+            } else {
+                // todo: should this throw?
+                return "null"
+            }
+        case .ArrayOfStrings:
```
schmit, line 182: note: for subject attributes this is not allowed in other SDKs. I know you coalesced the types so up to you how you want to handle it here

Sources/eppo/dto/EppoValue.swift
schmit: Not related to this PR, but wouldn't it make more sense to use the variation key instead of the variation value when hashing? Practically speaking wouldn't make a difference, but we avoid having to hash json blobs

Tests/eppo/EppoValueTests.swift
schmit: 👍 , we can always add more tests if we find more edge cases

schmit (approved):
🚀 so fast

Sources/eppo/dto/EppoValue.swift
leoromanovsky: If the instance attribute doubleValue is null then it will throw; the type is a throwable double so I need to handle the failure case.

Practically this won't happen since we are in the code branch for doubles which matches the implementation above.

I can either return a sentinel value here or throw an exception, but that would make this function a throwable string.

Sources/eppo/dto/EppoValue.swift
leoromanovsky: Yea good point I will review my old PR for why I decided to use this.

PR #31, opened 2024-06-12
Adds singleton instance and loadIfNeeded (FF-2401)

leoromanovsky:
## motivation

There are several behaviors that the SDK is not protecting against:

* Instantiating multiple instances of the `EppoClient`
* Performing network operations before each assignment

## description

* Adds a singleton pattern to the `EppoClient`
* Add ability to configure the `EppoClient` anywhere in an application with `configure`
* Replaces `load` method, which always performed a network request, with `loadIfNeeded`. Only loads a single time per application-lifecycle. Uses [swift actor pattern](https://www.swiftbysundell.com/articles/swift-actors/) to accomplish.

## question for reviewers

2. Does the readme for sdk usage make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Eppo iOS SDK
+# Eppo iOS (Swift) SDK
 
 [![Test and lint SDK](https://github.com/Eppo-exp/eppo-ios-sdk/actions/workflows/unit-tests.yml/badge.svg)](https://github.com/Eppo-exp/eppo-ios-sdk/actions/workflows/unit-tests.yml)
 
@@ -29,16 +29,16 @@ Begin by initializing a singleton instance of Eppo's client. Once initialized, t
 #### Initialize once
 
 ```swift
-var eppoClient: EppoClient = EppoClient("SDK-KEY-FROM-DASHBOARD");
+EppoClient.configure(apiKey: "SDK-KEY-FROM-DASHBOARD");
 ```
-
 
 #### Assign anywhere
 
 ```swift
 Task {
     do {
-        try await eppoClient.load();
+        let eppoClient = EppoClient.getInstance()
+        try await eppoClient.loadIfNeeded();
         self.assignment = try self.eppoClient.getStringAssignment(
             "new-user-onboarding"
             user.id,
@@ -57,14 +57,14 @@ For applications wrapping initialization and assignment in an `ObservableObject`
 
 ```swift
 @MainActor
-class AssignmentObserver : ObservableObject {
-    @Published var assignment: String?;
-    var eppoClient: EppoClient = EppoClient(EPPO-SDK-KEY);
+public class AssignmentObserver: ObservableObject {
+    @Published var assignment: String?
+    var eppoClient: EppoClient = EppoClient.configure(apiKey: EPPO_API_KEY)
 
     public init() {
         Task {
             do {
-                try await eppoClient.load();
+                try await eppoClient.loadIfNeeded()
                 self.assignment = try self.eppoClient.getStringAssignment(
                     "new-user-onboarding"
                     user.id,
@@ -72,8 +72,35 @@ class AssignmentObserver : ObservableObject {
                     "control"
                 );
             } catch {
-                self.assignment = nil;
+                self.assignment = nil
             }
+        }
+    }
+}
+```
+
+Rendering the view:
+
+```swift
+import SwiftUI
+
+struct ContentView: View {
+    @StateObject private var observer = AssignmentObserver()
+
+    var body: some View {
+        VStack {
+            if let assignment = observer.assignment {
+                Text("Assignment: \(assignment)")
+                    .font(.headline)
+                    .padding()
+            } else {
+                Text("Loading assignment...")
+                    .font(.subheadline)
+                    .padding()
+            }
+        }
+        .onAppear {
+            // You can perform additional actions on appear if needed
         }
     }
 }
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -57,28 +57,55 @@ For applications wrapping initialization and assignment in an `ObservableObject`
...
-    var eppoClient: EppoClient = EppoClient(EPPO-SDK-KEY);
+public class AssignmentObserver: ObservableObject {
+    @Published var assignment: String?
+    var eppoClient: EppoClient = EppoClient.configure(apiKey: EPPO_API_KEY)
 
     public init() {
         Task {
             do {
-                try await eppoClient.load();
+                try await eppoClient.loadIfNeeded()
```
leoromanovsky, line 67: Safe to call next to each assignment - once a single network request will happen.

Sources/eppo/EppoClient.swift
```diff
@@ -1,12 +1,13 @@
 import Foundation;
 
 // todo: make this a build argument (FF-1944)
-let sdkVersion = "1.2.1"
+let sdkVersion = "1.3.0"
```
leoromanovsky, line 4: This is a breaking API change and some behavioral changes but only subtly.

Tests/eppo/ClientTests.swift
```diff
@@ -108,307 +108,154 @@ final class eppoClientTests: XCTestCase {
...
-        try await eppoClient.load()
+        try await eppoClient.loadIfNeeded()
         
         _ = try eppoClient.getStringAssignment("6255e1a72a84e984aed55668", "randomization_algo")
         _ = try eppoClient.getStringAssignment("6255e1a72a84e984aed55668", "new_user_onboarding")
-
-        XCTAssertEqual(loggerSpy.logCount, 2, "Should log 2 times due to changing flags.")
-    }
-    
-    func testLoggingWhenRolloutIncreases() async throws {
```
leoromanovsky, line 249: It is not possible to execute the network request multiple times so this test does not make sense anymore.

Tests/eppo/ClientTests.swift
```diff
@@ -16,7 +16,7 @@ public class AssignmentLoggerSpy {
     var wasCalled = false
     var lastAssignment: Assignment?
     var logCount = 0
-
+
```
leoromanovsky, line 19: ran xcode "re-indent"

Sources/eppo/EppoClient.swift
```diff
@@ -18,34 +19,78 @@ public enum Errors: Error {
...
-        self.configurationStore = ConfigurationStore(requester: configurationRequester);
+        
+        let httpClient = NetworkEppoHttpClient(baseURL: host, apiKey: apiKey, sdkName: "sdkName", sdkVersion: "sdkVersion")
+        let configurationRequester = ConfigurationRequester(httpClient: httpClient)
+        self.configurationStore = ConfigurationStore(requester: configurationRequester)
     }
-
-    public func load() async throws {
+    
+    public static func configure(
```
aarsilv, line 63: we call this "init" in the other SDKs. Is "configure" a swift convention or just trying new lingo?

Tests/eppo/ClientTests.swift
aarsilv: 🧹 🧹 🧹

Tests/eppo/SingletonTest.swift
```diff
@@ -0,0 +1,48 @@
...
+                        try await eppoClient.loadIfNeeded()
+                        _ = try? eppoClient.getStringAssignment("subject_key", "some-assignment-key", [:])
+                        expectation.fulfill()
+                    }
+                }
+            }
+        }
+        
+        await fulfillment(of: [expectation], timeout: 5)
+        XCTAssertEqual(stubCallCount, 1)
```
aarsilv, line 46: :+1:

Sources/eppo/EppoClient.swift
felipecsl: FWIW, according to semver, breaking changes require bumping the major version, so this  should be v2.0.0

Tests/eppo/ClientTests.swift
felipecsl: are the trailing spaces intentional?

Sources/eppo/EppoClient.swift
felipecsl: +1 to keeping it consistent across languages

Tests/eppo/ClientTests.swift
leoromanovsky: XCode added that - I would like to bake https://github.com/apple/swift-format into this repo.

Screenshot 2024-06-12 at 9 51 52 AM

felipecsl: ugh i'm sorry you have to use xcode 😢

Sources/eppo/EppoClient.swift
leoromanovsky: We cannot use `init` as it is a reserved word in Swift for the class constructor: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/

Another note: the `configure` method while it sets the desired user options does NOT load the RAC/UFC data. Our documentation suggestions callers wrap `loadIfDesired` in a `Task` but I will investigate if this is feasible to happen within the SDK.

Sources/eppo/EppoClient.swift
leoromanovsky: Yes agreed - I was wondering if we could fudge it a bit but thanks for lending your vote, will bump to 2.0.0. We will be releasing UFC as 3.0.0 and include these changes.

Tests/eppo/ClientTests.swift
leoromanovsky: 🤣 I'm using xcode for the debugger but mostly writing code in cursor to have the ai assist. surprisingly it is working pretty well together! both programs reload from disk often.

PR #33, opened 2024-06-17
Fix: [assignment cache is sensitive to changes in allocation or variation key] (FF-2435)

leoromanovsky:
## Observed behavior

The assignment cache is not logging changes in allocations.

## Desired behavior

The assignment cache should log the assignment if either the allocation or variation key change.

## Details

A caveat: the variation key is not available in the RAC. This change is being developed against the `1.x` branch in order to fix an undesirable bug. A modification will be developed against the `3.x` UFC branch. As such as are using the stringified variation value as a substitute.

* Converts the `Map[String, String]` to a `Map[CacheKey, CacheValue]` - this allows us to pick the two attributes from the public interface `AssignmentCacheKey` which break the cache when they change. In this change that is being expanded from just the variation value to the allocation key too. With the change away from `String`, some code can be removed by implementing the `Hashable` and `Equatable` interfaces in the `CacheKey` and `CacheValue`, respectively.
* Keeping the interface to the class the same.

## Deployment

* Tag `1.2.2` will be created from this branch and a release created for it.
* An additional branch will be made against (https://github.com/Eppo-exp/eppo-ios-sdk/pull/21) to use the actual variation key.

### DIFF
--- a/Tests/eppo/AssignmentCacheTests.swift
+++ b/Tests/eppo/AssignmentCacheTests.swift
@@ -15,33 +15,66 @@ final class AssignmentCacheTests: XCTestCase {
         cache = nil
         super.tearDown()
     }
-    
-    func testSetAndGet() {
-        let key = "testKey"
-        let value = "testValue"
-        XCTAssertNil(cache.get(key: key), "Cache should return nil for a key that has not been set.")
-        cache.set(key: key, value: value)
-        XCTAssertEqual(cache.get(key: key), value, "Cache should return the value that was set for a key.")
-    }
-    
-    func testHas() {
-        let key = "testKey"
-        XCTAssertFalse(cache.has(key: key), "Cache should return false for a key that has not been set.")
-        cache.set(key: key, value: "testValue")
-        XCTAssertTrue(cache.has(key: key), "Cache should return true for a key that has been set.")
-    }
-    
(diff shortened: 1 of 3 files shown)


Tests/eppo/AssignmentCacheTests.swift
```diff
@@ -15,33 +15,66 @@ final class AssignmentCacheTests: XCTestCase {
...
+        let key2 = AssignmentCacheKey(
+            subjectKey: "Math", 
+            flagKey: "TestFlag",
+            allocationKey: "A2", // changes
+            variationKey: EppoValue(value: "VariationA", type: EppoValueType.String).toHashedString()
+        )
+
+        cache.setLastLoggedAssignment(key: key1)
+        XCTAssertTrue(cache.hasLoggedAssignment(key: key1))
+        XCTAssertFalse(cache.hasLoggedAssignment(key: key2))
```
leoromanovsky, line 35: This assertion is a stand-in for whether the event would be logged. Do you think it is sufficient to test this behavior?

felipecsl: LGTM

felipecsl (approved):
👍🏽 looks pretty similar to what I did for JS 💯

PR #35, opened 2024-06-18
Unified interface into a single method `initialization`

leoromanovsky:
## Observed behavior

In (https://github.com/Eppo-exp/eppo-ios-sdk/pull/31) we added support for a singleton client but the method to configure and fetch flag configuration remained separated.

## Desired behavior

Mirror the behavior of our other SDKs to have a single method to configure and fetch flag configurations.

## Changes

* Unify into a single async `initialize` method. It returns the instance with loaded flag configurations and can be called multiple times. The existing `checkAndSetLoaded` actor pattern makes sure only a single network request is made.
* Use the static method `shared` to return the singleton; this seems to be the canonical naming in Swift.
* Create a new instance if initialization is requested with a different SDK key or host.
* New unit tests to demonstrate behavior.

## For reviewers

See the updated README for examples of usage and if it makes sense from the user perspective.

## Deployment

This is being merged to `main` in the `3.x` branch.


README.md
```diff
@@ -24,49 +24,44 @@ While in XCode:
...
             do {
-                try await eppoClient.loadIfNeeded()
-                self.assignment = try self.eppoClient.getStringAssignment(
-                    "new-user-onboarding"
+                let eppoClient = try await EppoClient.initialize(apiKey: EPPO_API_KEY)
+                self.assignment = try eppoClient.getStringAssignment(
+                    "new-user-onboarding",
                     user.id,
                     user.attributes,
                     "control"
```
leoromanovsky, line 67: Here is how most will use this - there will be several of these `Observer` classes. Does it seem odd to be calling `EppoClient.initialize` or it is obvious that it's a singleton? Could call it `initializeSingletonIfNeeded` or something obvious - thanks.

README.md
```diff
@@ -24,49 +24,44 @@ While in XCode:
...
 It is recommended to wrap initialization in a `Task` block in order to perform network request asynchronously.
 
-For applications wrapping initialization and assignment in an `ObservableObject` is recommended. This will create an object that will update Swift UI when the assignment is received.
+For applications wrapping initialization and assignment in an `ObservableObject` is the best practice. This will create an object that will update Swift UI when the assignment is received.
 
 ```swift
 @MainActor
 public class AssignmentObserver: ObservableObject {
     @Published var assignment: String?
-    var eppoClient: EppoClient = EppoClient.configure(apiKey: EPPO_API_KEY)
```
leoromanovsky, line 62: Cannot call `try await EppoClient.initialize(apiKey: "test")` here as this is a property initializer.

leoromanovsky:
I am reading that the `shared` naming is more canonical in Swift:

* https://developer.apple.com/documentation/swift/managing-a-shared-resource-using-a-singleton
* https://www.hackingwithswift.com/example-code/language/what-is-a-singleton
* https://cocoacasts.com/what-is-a-singleton-and-how-to-create-one-in-swift
* https://medium.com/@nimjea/singleton-class-in-swift-17eef2d01d88

Let me make those changes.

Sources/eppo/EppoClient.swift
```diff
@@ -60,37 +60,48 @@ public class EppoClient {
...
+        guard let instance = sharedInstance else {
             throw Errors.notConfigured
         }
         return instance
     }
     
-    public static func resetInstance() {
-        instance = nil
+    public static func resetSharedInstance() {
+        sharedInstance = nil
```
felipecsl, line 98: is this only for tests? if do I'd recommend leaving a comment to explicitly point that out since this could have dangerous consequences in prod

leoromanovsky: No - since this is a singleton we are allowing people to call this method repeatedly. Also since this is an async method people might choose to call it from app start, but there would be a race condition with the observer. Therefore they might want to put initialization next to assignment and this init will be called a lot. The readme has both of these examples.

Sources/eppo/EppoClient.swift
```diff
@@ -31,10 +31,10 @@ actor EppoClientState {
     }
 }
 
-public class EppoClient {
+public class EppoClient : Equatable {
     public typealias AssignmentLogger = (Assignment) -> Void
     
-    private static var instance: EppoClient?
+    private static var sharedInstance: EppoClient?
```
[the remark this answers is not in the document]
leoromanovsky: Thanks for checking as well - that's basically how I arrived at this pattern too. Apple's doc don't explicit say to use "shared" but theirs and many libraries use it.

Sources/eppo/EppoClient.swift
```diff
@@ -60,37 +60,48 @@ public class EppoClient {
...
-            instance = EppoClient(apiKey: apiKey, host: host, assignmentLogger: assignmentLogger, assignmentCache: assignmentCache)
+    ) async throws -> EppoClient {
+        let tentativeNewInstance = EppoClient(
+            apiKey: apiKey,
+            host: host,
+            assignmentLogger: assignmentLogger, 
+            assignmentCache: assignmentCache
+        )
+
+        if let instance = sharedInstance, instance != tentativeNewInstance {
```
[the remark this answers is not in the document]
leoromanovsky: sharedInstance is an optional and may be null

leoromanovsky:
* Removed the feature which re-created the singleton if the api key or host changed
* changed `apiKey` to `sdkKey` throughout, except the network request as the server has not been migrated
* Introduced more robust code for creating the singleton and running the network request. Tested with 50 concurrent threads creating singletons and requesting the configuration.

Tests/eppo/SingletonTest.swift
```diff
@@ -21,28 +22,42 @@ final class EppoTests: XCTestCase {
...
             await withThrowingTaskGroup(of: Void.self) { group in
                 for _ in 0 ..< expectedCount {
                     group.addTask {
-                        try await eppoClient.loadIfNeeded()
-                        _ = try? eppoClient.getStringAssignment(flagKey: "some-assignment-key", subjectKey: "subject_key", subjectAttributes: [:], defaultValue: "default")
+                        _ = try await EppoClient.initialize(sdkKey: "mock-api-key")
+                        _ = try? EppoClient.shared().getStringAssignment(flagKey: "some-assignment-key", subjectKey: "subject_key", subjectAttributes: [:], defaultValue: "default")
                         expectation.fulfill()
                     }
                 }
```
leoromanovsky, line 63: This existing test run 50 concurrent threads which all attempt to initialize the client, get a singleton and perform assignment. The expectation is that only 1 network request takes place.

Tests/eppo/SingletonTest.swift
```diff
@@ -21,28 +22,42 @@ final class EppoTests: XCTestCase {
...
+        let eppoClient1 = try await EppoClient.initialize(sdkKey: "mock-api-key1")
+        let eppoClient2 = try await EppoClient.initialize(sdkKey: "mock-api-key2")
+        XCTAssertIdentical(eppoClient1, eppoClient2, "Changing SDK key does not re-instantiate the singleton")
+    }
+    
+    func testInitializeWithDifferentHost() async throws {
+        let eppoClient1 = try await EppoClient.initialize(sdkKey: "mock-api-key1")
+        let eppoClient2 = try await EppoClient.initialize(sdkKey: "mock-api-key1", host: "https://test.cloud")
+        XCTAssertIdentical(eppoClient1, eppoClient2, "Changing host re-instantiates the singleton")
     }
```
leoromanovsky, line 48: Not changes in configuration force a new instance

Sources/eppo/EppoClient.swift
```diff
@@ -47,52 +48,66 @@ public class EppoClient {
...
+                            continuation.resume(returning: sharedInstance!)
+                        } catch {
+                            continuation.resume(throwing: error)
+                        }
+                    }
+                } else {
+                    continuation.resume(returning: sharedInstance!)
+                }
+            }
         }
```
leoromanovsky, line 93: This is quite complex and I arrived at it when needing only a single thread to access the critical section, hence `barrier`. Inside if the instance is not created, create one. Then start an SDK owned Task which performs the configuration fetching in the background.

leoromanovsky: chatgpt explains it well - https://chatgpt.com/share/550bea2f-86b2-4c75-83d0-8c18cc9325a9

felipecsl (approved):

Tests/eppo/SingletonTest.swift
aarsilv: 🔥

aarsilv (approved):

PR #41, opened 2024-09-14
feat: [add support for initializing the SDK from a static json config…

leoromanovsky:
…uration] (FF-3242)

## Customer requirement

1. Provide a json UFC configuration to the SDK and synchronously put it into a mode available for assignment. This is useful for when the configuration is downloaded from an external process and passed to the mobile device over an internal API. Alternatively it can be the pass that a UFC configuration is packaged with the local build.
2. After initializing with a static configuration, users want to (optionally) be able to fetch configuration from the remote CDN.

**Design question**

The ability to set static configuration is separated from loading the remote configuration. The main benefit here is for the former to be a sync method and make the SDK immediately available.

Customers might want to then follow up by loading a remote configuration if the packaged configuration is stale. Others who provide the configuration over their internal API from a server process will be satisfied with it.

The only re-render should occur after the remote configuration is loaded.

## Changes

Adds a `initialize` method which is synchronous and takes as its inputs a json string of the UFC configuration, whether the format is obfuscated and the usual assignment logger and assignment cache.

The `ConfigurationRequester` is refactored into an interface (`protocol` in swift) and a new implementation handles inputing the json string and decoding it from JSON. The [json decode ](https://github.com/Eppo-exp/eppo-ios-sdk/blob/main/Sources/eppo/UniversalFlagConfig.swift#L24-L42)is already well tested with appropriate error messages for various errors.

During sync init, the configuration store is set atomically and safely (using existing methods) and the SDK put into `offline mode`, which means the SDK key and host parameter is are not validated to be set.

## Follow-ups

* Add a persistent cache to the swift sdk

## Testing

* Refactored the data tests to execute both initialization from a network request (stubbed) or from json string.
* demonstrates that a configuration can be sync loaded, assignment working, remote reloaded and assignment updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,42 @@ Task {
 }
 ```
 
+#### Offline initialization
+
+If you'd like to initialize Eppo's client without performing a network request, you can pass in a pre-fetched configuration JSON string.
+
+```swift
+try EppoClient.initialize(
+    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
+    obfuscated: false
+);
+```
+
+The `obfuscated` parameter is used to inform the SDK if the flag configuration is obfuscated.
+
+The initialization method is synchronous and allows you to perform assignments immediately.
+
+**(Optional) Fetching the configuration from the remote source on-demand.**
+
+Combine both initialization methods to start your application with a local configuration and then 
+asynchronously fetch the latest flag configuration from the remote source.
+
+```swift
+try EppoClient.initialize(
+    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
+    obfuscated: false,
+);
+
+...
+
+Task {
+    try await EppoClient.initialize(
+        sdkKey: "SDK-KEY-FROM-DASHBOARD",
+        forceReinitialize: true
+    );
+}
+```
+
 #### Assign anywhere
 
 Assignment is a synchronous operation.
--- a/README.md
+++ b/README.md
@@ -46,8 +46,9 @@ If you'd like to initialize Eppo's client without performing a network request, 
 
 ```swift
 try EppoClient.initialize(
-    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
-    obfuscated: false
+    configurationJson: "{ \"pre-loaded-JSON\": \"passed in here\" }",
+    obfuscated: false,
+    sdkKey: "SDK-KEY-FROM-DASHBOARD"
 );
 ```
 
@@ -55,24 +56,22 @@ The `obfuscated` parameter is used to inform the SDK if the flag configuration i
 
 The initialization method is synchronous and allows you to perform assignments immediately.
 
-**(Optional) Fetching the configuration from the remote source on-demand.**
+#### (Optional) Fetching the configuration from the remote source on-demand.**
 
 Combine both initialization methods to start your application with a local configuration and then 
 asynchronously fetch the latest flag configuration from the remote source.
 
 ```swift
 try EppoClient.initialize(
-    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
+    configurationJson: "{ \"pre-loaded-JSON\": \"passed in here\" }",
     obfuscated: false,
+    sdkKey: "SDK-KEY-FROM-DASHBOARD",
 );
 
 ...
 
 Task {
-    try await EppoClient.initialize(
-        sdkKey: "SDK-KEY-FROM-DASHBOARD",
-        forceReinitialize: true
-    );
+    try await EppoClient.shared().load();
 }
 ```
 
--- a/Sources/eppo/ConfigurationRequester.swift
+++ b/Sources/eppo/ConfigurationRequester.swift
@@ -2,7 +2,11 @@ import Foundation;
 
 let UFC_CONFIG_URL = "/api/flag-config/v1/config"
 
-class ConfigurationRequester {
+protocol ConfigurationRequesterProtocol {
+    func fetchConfigurations() async throws -> UniversalFlagConfig
+}
+
+class HttpConfigurationRequester: ConfigurationRequesterProtocol {
     private let httpClient: EppoHttpClient;
 
     public init(httpClient: EppoHttpClient) {
@@ -12,5 +16,17 @@ class ConfigurationRequester {
     public func fetchConfigurations() async throws -> UniversalFlagConfig {
         let (urlData, _) = try await httpClient.get(UFC_CONFIG_URL);
         return try UniversalFlagConfig.decodeFromJSON(from: String(data: urlData, encoding: .utf8)!);
+    }
+}
+
+class JsonConfigurationRequester: ConfigurationRequesterProtocol {
+    private let configurationJson: String;
+
+    public init(configurationJson: String) {
+        self.configurationJson = configurationJson
+    }
+
+    public func fetchConfigurations() throws -> UniversalFlagConfig {
+        return try UniversalFlagConfig.decodeFromJSON(from: configurationJson);
     }
 }
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -4,6 +4,8 @@ import Foundation;
 public let sdkName = "ios"
 public let sdkVersion = "3.2.0"
 
+public let default_host = "https://fscdn.eppo.cloud"
+
 public enum Errors: Error {
     case notConfigured
     case sdkKeyInvalid
@@ -38,7 +40,6 @@ public class EppoClient {
     
     private var flagEvaluator: FlagEvaluator = FlagEvaluator(sharder: MD5Sharder())
     private(set) var isConfigObfuscated = true;
-    private(set) var isOnlineMode = false
     
     private(set) var sdkKey: String
     private(set) var host: String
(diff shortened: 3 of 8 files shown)


Sources/eppo/EppoClient.swift
```diff
@@ -225,11 +260,11 @@ public class EppoClient {
         subjectAttributes: SubjectAttributes,
         expectedVariationType: UFC_VariationType) throws -> FlagEvaluation?
     {
-        if (self.sdkKey.count == 0) {
+        if (isOnlineMode && self.sdkKey.count == 0) {
```
leoromanovsky, line 263: Note to self: This behavior needs a test

Sources/eppo/EppoClient.swift
```diff
@@ -51,32 +52,41 @@ public class EppoClient {
...
         self.configurationStore = ConfigurationStore(requester: configurationRequester)
     }
     
     public static func initialize(
         sdkKey: String,
         host: String = "https://fscdn.eppo.cloud",
         assignmentLogger: AssignmentLogger? = nil,
-        assignmentCache: AssignmentCache? = InMemoryAssignmentCache()
+        assignmentCache: AssignmentCache? = InMemoryAssignmentCache(),
+        forceReinitialize: Bool = false
```
leoromanovsky, line 71: This method respects the singleton. For the use case of wanting to load the remote configuration after the static one, need to be able to create a new singleton with the sdk key.

leoromanovsky: 🔴 after another read I suspect there is an undesirable behavior for customers that perform a remote load after setting a static configuration in that the newly created singleton will not have remote configuration loaded and be blank for a short period of time.

I will take another pass at this implementation to cover it.

leoromanovsky: What needs to happen here is the configuration be loaded from remote and only then the singleton be over-written.

Note to self: write a test for this case.

README.md
```diff
@@ -40,6 +40,42 @@ Task {
 }
 ```
 
+#### Offline initialization
+
+If you'd like to initialize Eppo's client without performing a network request, you can pass in a pre-fetched configuration JSON string.
```
leoromanovsky, line 45: Note to self: mention that if using a packaged and stale configuration it might not be desirable to add an assignment logger to the client; for this case better to do so when initializing from remote configuration.

For others who are passing a configuration over the wire they are likely satisfied with an application logger at this step.

aarsilv (commented):
Great stuff! Love the analogous configuration requester for JSON string.

My main thought is instead of requiring another initialization, we should make it an option if we'll attempt a fetch from CDN. Then we'll have just a single instance the whole time and no re-initialization period

README.md
```diff
@@ -40,6 +40,42 @@ Task {
...
+    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
+    obfuscated: false
+);
+```
+
+The `obfuscated` parameter is used to inform the SDK if the flag configuration is obfuscated.
+
+The initialization method is synchronous and allows you to perform assignments immediately.
+
+**(Optional) Fetching the configuration from the remote source on-demand.**
```
rasendubi, line 58: nit: is this supposed to be a heading?
```suggestion
##### (Optional) Fetching the configuration from the remote source on-demand.
```

README.md
```diff
@@ -40,6 +40,42 @@ Task {
...
+);
+
+...
+
+Task {
+    try await EppoClient.initialize(
+        sdkKey: "SDK-KEY-FROM-DASHBOARD",
+        forceReinitialize: true
+    );
+}
```
rasendubi, line 76: This API is super-confusing

From the code alone, I would expect the second call to reset and loose initial configuration from the previous call. `forceReinitialize: true` also reinforces this expectation.

Sources/eppo/EppoClient.swift
rasendubi: I support what @aarsilv is saying.

I would say that we must preserve `EppoClient` identity in case the result of the first `initialize()` call is saved by the user. Calling `initialize` twice is very confusing as well.

I have implemented this feature recently in Python with an optional `initial_configuration` parameter: https://github.com/Eppo-exp/python-sdk/pull/70
I also suggest adding a similar `Configuration` abstraction as it's very likely that we'll allow users to initialize bandit configuration in the future as well.

Sources/eppo/EppoClient.swift
```diff
@@ -117,6 +151,7 @@ public class EppoClient {
         guard !alreadyLoaded else { return }
         
         try await self.configurationStore.fetchAndStoreConfigurations()
+        self.isOnlineMode = true
```
rasendubi, line 154: 🐛  `loadIfNeeded` is called in an async context and `isOnlineMode` is set only after configuration is successfully fetched and stored, so the user might call assignment functions before `isOnlineMode` is set.

This nullifies the checks in `getInternalAssignment()` because if `isOnlineMode` is `true`, that means that we fetched the configuration successfully, and therefore api key and host are valid.

Sources/eppo/EppoClient.swift
rasendubi: I think it's fine to request an sdk key even if client is initialized from json. We want them to hold of a valid SDK key, and we might use it for something else in the future.

Sources/eppo/ConfigurationRequester.swift
```diff
@@ -14,3 +18,15 @@ class ConfigurationRequester {
...
+class JsonConfigurationRequester: ConfigurationRequesterProtocol {
+    private let configurationJson: String;
+
+    public init(configurationJson: String) {
+        self.configurationJson = configurationJson
+    }
+
+    public func fetchConfigurations() throws -> UniversalFlagConfig {
+        return try UniversalFlagConfig.decodeFromJSON(from: configurationJson);
+    }
```
rasendubi, line 31: minor: this will trigger JSON-decoding on every call to `fetchConfigurations()` instead of only once

rasendubi, line 31: minor: I was looking to see if this would get triggered every 30 seconds by configuration poller and it looks like iOS SDK doesn't have the poller at all. Is this intentional? If so, we should update the README as it says the SDK does poll the configuration every 30s

Sources/eppo/EppoClient.swift
rasendubi, line 71: minor: `EppoClient` exposes a `resetSharedInstance()`, which does the same but is more explicit about what's happening

README.md
rasendubi: UPD: found that it does indeed create a new client. This is problematic because:
- the new client does not have the configuration passed in JSON, so in this example, it would still be returning default assignments as if no json configuration were passed
- this invalidates the client returned from the first `initialize`. If user save that client, they must know to re-store the new reference. Otherwise, they would continue getting assignments from JSON.
- re-initialization creates a second assignment cache that is not shared with the first one

Sources/eppo/EppoClient.swift
```diff
@@ -92,6 +102,30 @@ public class EppoClient {
             }
         }
     }
+
+    public static func initialize(
+        configurationJson: String,
+        obfuscated: Bool,
+        assignmentLogger: AssignmentLogger? = nil,
+        assignmentCache: AssignmentCache? = InMemoryAssignmentCache()
+    ) throws -> EppoClient {
```
rasendubi, line 111: minor: `EppoClient`'s `load()` and `loadIfNeeded()` method behave strangely/unexpectedly when client is initialized with this method.

README.md
rasendubi, line 76: minor: after looking at `EppoClient` API, it looks like the best interface for this use case would be similar to this:
```suggestion
try EppoClient.initialize(
    sdkKey: "SDK-KEY-FROM-DASHBOARD",
    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
    obfuscated: false,
);

...

Task {
    try await EppoClient.load();
}
```

README.md
```diff
@@ -40,6 +40,42 @@ Task {
...
+If you'd like to initialize Eppo's client without performing a network request, you can pass in a pre-fetched configuration JSON string.
+
+```swift
+try EppoClient.initialize(
+    configurationJson: "PRE-FETCHED-CONFIGURATION-JSON",
+    obfuscated: false
+);
+```
+
+The `obfuscated` parameter is used to inform the SDK if the flag configuration is obfuscated.
```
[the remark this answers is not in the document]
leoromanovsky: Yes - the client can also guess that the configuration is obfuscated by inspecting it with reasonable accuracy.

Sources/eppo/ConfigurationRequester.swift
leoromanovsky: That's correct there is no included poller. Ok good idea I'll make a note.

Sources/eppo/EppoClient.swift
```diff
@@ -105,7 +151,22 @@ public class EppoClient {
...
+        sdkKey: String,
+        host: String = default_host
+    ) {
+        let httpClient = NetworkEppoHttpClient(
+            baseURL: host,
+            sdkKey: sdkKey,
+            sdkName: sdkName,
+            sdkVersion: sdkVersion
+        )
+        self.isConfigObfuscated = true
```
rasendubi, line 168: 🐛 this changes config obfuscation _before_ new configuration have been fetched, which will make the client re-interpret previously supplied configuration as obfuscated even if it is not (until the client successfully fetches new configuration)

PR #42, opened 2024-09-16
FF-3242 feat: support passing initial configuration and offline initialization

rasendubi:
This adds support for passing initial configuration in initialize and offline initialization method. In a sense, they are two orthogonal features often used together.

Overview of the changes:
- Decouple `ConfigurationStore` from `ConfigurationRequester`
- Introduce `Configuration` type that encapsulates everything we need to know about configuration: flags configuration and whether it's obfuscated (+ will add bandits configuration in the future). `ConfigurationStore` now just stores a single instance of configuration atomically.
- Add `initializeOffline` method that works the same as regular initialization but does not request a configuration
- Add `initialConfiguration` parameter to initialization methods. If passed, it initializes configuration store with that value
- Fix data race accessing `sharedInstance` (reads were not properly guarded)

PR #43, opened 2024-09-16
feat: persistent configuration to disk (FF-2243)

leoromanovsky:
## Motivation

* When the SDK is initialized, it might not have network access. This is particularly true on mobile. Users of the SDK want to the last downloaded configuration to be cached on disk so they can provide the same UX until network becomes available again.

## Changes

Augment the `setConfiguration` method to update disk as well as memory.

Writing to disk is a fairly low-level interface but straight-forward since we are treating the configuration as a blob and not interested in loading from it, so a database is not needed.

The synchronization on the method is [DispatchQueue.asyncAndWait](https://developer.apple.com/documentation/dispatch/dispatchqueue/3191901-asyncandwait), which is blocking but is called from an async method:

```
    // Loads the configuration from the remote source on-demand. Can be used to refresh as desired.
    //
    // This function can be called from multiple threads; synchronization is provided to safely update
    // the configuration cache but each invocation will execute a new network request with billing impact.
    public func load() async throws {
        let config = try await self.configurationRequester.fetchConfigurations()
        self.configurationStore.setConfiguration(configuration: config)
    }
```

Configuration is loaded from disk on class init, which will slow down `EppoClient` initialize. In the case of `initialConfiguration` provided there will be an additional write as well.

No changes are made to the read (assignment path), this continues to be served only from memory.

### DIFF
--- a/Sources/eppo/Configuration.swift
+++ b/Sources/eppo/Configuration.swift
@@ -2,14 +2,20 @@ import Foundation
 
 public struct Configuration {
     internal let flagsConfiguration: UniversalFlagConfig;
+    internal let obfuscated: Bool;
 
 
-    internal init(flagsConfiguration: UniversalFlagConfig) {
+    internal init(flagsConfiguration: UniversalFlagConfig, obfuscated: Bool) {
         self.flagsConfiguration = flagsConfiguration
+        self.obfuscated = obfuscated
     }
 
-    public init(flagsConfigurationJson: Data) throws {
+    public init(flagsConfigurationJson: Data, obfuscated: Bool) throws {
         let flagsConfiguration = try UniversalFlagConfig.decodeFromJSON(from: flagsConfigurationJson);
-        self.init(flagsConfiguration: flagsConfiguration)
+        self.init(flagsConfiguration: flagsConfiguration, obfuscated: obfuscated)
+    }
+
+    internal func getFlag(flagKey: String) -> UFC_Flag? {
+        return self.flagsConfiguration.flags[flagKey]
     }
 }
--- a/Sources/eppo/Configuration.swift
+++ b/Sources/eppo/Configuration.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-public struct Configuration {
+public struct Configuration: Codable {
     internal let flagsConfiguration: UniversalFlagConfig;
     internal let obfuscated: Bool;
 
--- a/Sources/eppo/Configuration.swift
+++ b/Sources/eppo/Configuration.swift
@@ -1,0 +1,15 @@
+import Foundation
+
+public struct Configuration {
+    internal let flagsConfiguration: UniversalFlagConfig;
+
+
+    internal init(flagsConfiguration: UniversalFlagConfig) {
+        self.flagsConfiguration = flagsConfiguration
+    }
+
+    public init(flagsConfigurationJson: Data) throws {
+        let flagsConfiguration = try UniversalFlagConfig.decodeFromJSON(from: flagsConfigurationJson);
+        self.init(flagsConfiguration: flagsConfiguration)
+    }
+}
--- a/Sources/eppo/ConfigurationStore.swift
+++ b/Sources/eppo/ConfigurationStore.swift
@@ -1,32 +1,54 @@
 import Foundation
 
-private let fileName = "eppo-configuration.json"
-
 class ConfigurationStore {
     private var configuration: Configuration?
     private let syncQueue = DispatchQueue(
         label: "com.eppo.configurationStoreQueue", attributes: .concurrent)
-    
-    private let fileURL: URL
+
+    private let cacheFileURL: URL?
+    // This is a serial (non-concurrent) queue, so writers don't fight
+    // each other and last writer wins.
+    //
+    // The queue is static because if there are multiple stores, they
+    // would be sharing the cache file.
+    static let persistenceQueue = DispatchQueue(
+      label: "com.eppo.configurationStorePersistence", qos: .background)
     
     // Initialize with the disk-based path for storage
-    public init() {
-        guard let cacheDirectory = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first else {
-            fatalError("Unable to access cache directory")
-        }
-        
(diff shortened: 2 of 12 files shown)


Sources/eppo/Configuration.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-public struct Configuration {
+public struct Configuration: Codable {
```
leoromanovsky, line 3: `Codable` interface is for converting objects to and from json.

Sources/eppo/ConfigurationStore.swift
```diff
@@ -2,11 +2,20 @@ import Foundation
...
 class ConfigurationStore {
     private var configuration: Configuration?
-    
     private let syncQueue = DispatchQueue(
         label: "com.eppo.configurationStoreQueue", attributes: .concurrent)
     
+    private let fileURL: URL
+    
+    // Initialize with the disk-based path for storage
     public init() {
```
typotter, line 11: What do you think about passing through the `initialConfiguration` optional parameter available from [`EppoClient`](https://github.com/Eppo-exp/eppo-ios-sdk/pull/42/files#diff-dc60401cf877df4122e1cb0b73c011aba7a9a665785dc943c9855d104de4d7e9R69)? We can skip the initial disk read when there's a configuration string.

Sources/eppo/dto/EppoValue.swift
```diff
@@ -188,3 +188,20 @@ public class EppoValue : Decodable, Equatable {
         }
     }
 }
+extension EppoValue {
```
typotter, line 191: Is this to support the `Codable` interface?

typotter (approved):
Looking good! Just one suggestion to consider.

Sources/eppo/dto/EppoValue.swift
leoromanovsky: Yes

Sources/eppo/ConfigurationStore.swift
leoromanovsky: Good idea!

aarsilv (approved):
Nice job getting this in too!

Tests/eppo/ConfigurationStoreTests.swift
```diff
@@ -30,6 +33,24 @@ final class ConfigurationStoreTests: XCTestCase {
...
     
+    override func tearDownWithError() throws {
+        try super.tearDownWithError()
+        // Clean up the configuration file
+        cleanUpConfigurationFile()
+        // Release the reference to ensure it's deallocated
+        configurationStore = nil
+    }
+    
+    private func cleanUpConfigurationFile() {
```
rasendubi, line 44: nit: does it make sense to add this to `ConfigurationStore`? so the test doesn't need to know where and how configuration is stored

Tests/eppo/ConfigurationStoreTests.swift
```diff
@@ -30,6 +33,24 @@ final class ConfigurationStoreTests: XCTestCase {
...
+        let fileManager = FileManager.default
+        let urls = fileManager.urls(for: .documentDirectory, in: .userDomainMask)
+        let fileURL = urls[0].appendingPathComponent("configuration.json")
+        
+        if fileManager.fileExists(atPath: fileURL.path) {
+            try? fileManager.removeItem(at: fileURL)
+        }
```
rasendubi, line 51: nit: no need to check for existence if we're going to delete the file anyway
```suggestion
        try? fileManager.removeItem(at: fileURL)
```

Sources/eppo/ConfigurationStore.swift
```diff
@@ -2,11 +2,20 @@ import Foundation
...
+        // Set the file path for storing the configuration
+        let fileManager = FileManager.default
+        let urls = fileManager.urls(for: .documentDirectory, in: .userDomainMask)
+        self.fileURL = urls[0].appendingPathComponent("configuration.json")
+        
+        // Load any existing configuration from disk when initializing
+        self.configuration = loadFromDisk()
```
rasendubi, line 18: minor: would it make sense to spawn this concurrently, so the store is returned immediately and configuration is populated shortly after?

Sources/eppo/ConfigurationStore.swift
```diff
@@ -2,11 +2,20 @@ import Foundation
 
 class ConfigurationStore {
     private var configuration: Configuration?
-    
     private let syncQueue = DispatchQueue(
         label: "com.eppo.configurationStoreQueue", attributes: .concurrent)
     
+    private let fileURL: URL
```
rasendubi, line 8: minor: Adding persistence to `ConfigurationStore` seems obvious. I'm not sure if that's the best place though. One concern I have is that it adds another responsibility to the store and couples persistence with managing global configuration (so they would be harder to separate in the future). e.g., one place where it shows up is complicating tests with the cleanup step

Questions to think about:
- is configuration caching something that users would want to ever disable?
- is it something that users would want to configure? could there be different sensible strategies?

But I'm probably over-complicating this, so feel free to ignore

rasendubi (approved):

Sources/eppo/ConfigurationStore.swift
```diff
@@ -25,6 +34,30 @@ class ConfigurationStore {
...
+    }
+    
+    // Save the configuration to disk
+    private func saveToDisk(configuration: Configuration) {
+        do {
+            let data = try JSONEncoder().encode(configuration)
+            try data.write(to: fileURL)
+        } catch {
+            print("Error saving configuration to disk: \(error)")
+        }
```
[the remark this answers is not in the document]
leoromanovsky: Thanks - I will research this.

Sources/eppo/ConfigurationStore.swift
leoromanovsky: It's a good thought for sure. There are some downsides, particularly the disk performance latency. Let me consider making this a default and allowing users to disable it.

rasendubi: If we factor configuration store into interface, the disk could be an adapter/wrapper (similar to caching assignment logger in python)

PR #44, opened 2024-09-18
publishing instructions

leoromanovsky:
This deviates from other repositories. Are the instructions clear enough?


aarsilv (approved):
Makes sense to me! I wonder (for later) if we can add in some protection to fail the release if the `v` is there.

rasendubi (approved):
Looks good to me

PR #45, opened 2024-09-19
#43 improvements

rasendubi:
1. Make caching optional and handle the case of absent / non-writable cache directory
2. Move cache saving to background, so it does not slow configuration assignment
3. Make `clearPersistentCache()` not clear configuration. From a user perspective, I would expect this function to only clear the cache

### DIFF
--- a/Sources/eppo/ConfigurationStore.swift
+++ b/Sources/eppo/ConfigurationStore.swift
@@ -1,32 +1,47 @@
 import Foundation
 
-private let fileName = "eppo-configuration.json"
-
 class ConfigurationStore {
     private var configuration: Configuration?
     private let syncQueue = DispatchQueue(
         label: "com.eppo.configurationStoreQueue", attributes: .concurrent)
     
-    private let fileURL: URL
+    private let cacheFileURL: URL?
     
     // Initialize with the disk-based path for storage
-    public init() {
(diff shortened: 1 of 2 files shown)


Sources/eppo/ConfigurationStore.swift
```diff
@@ -1,32 +1,54 @@
...
         label: "com.eppo.configurationStoreQueue", attributes: .concurrent)
-    
-    private let fileURL: URL
+
+    private let cacheFileURL: URL?
+    // This is a serial (non-concurrent) queue, so writers don't fight
+    // each other and last writer wins.
+    //
+    // The queue is static because if there are multiple stores, they
+    // would be sharing the cache file.
```
aarsilv, line 13: 👍

Sources/eppo/ConfigurationStore.swift
```diff
@@ -44,38 +66,50 @@ class ConfigurationStore {
...
-                try FileManager.default.removeItem(at: self.fileURL)
+                try FileManager.default.removeItem(at: cacheFileURL)
             } catch {
                 print("Error removing cache file: \(error)")
             }
         }
     }
     
-    // Save the configuration to disk
+    // Save the configuration to disk (in background)
```
aarsilv, line 88: 🔥

aarsilv (approved):
Changes are looking good from my end!
(Disclaimer: I don't know Swift 😅 )

leoromanovsky (approved):

Issue #46, opened 2024-10-01
Crash on iOS 15 


alrocha:
Hi eppo team, 

I am opening this issue due to a crash encountered from version 3.1.0 in your repository compiling the app with iOS 15. With 3.0.1 everything works fine. 

```
dyld[3124]: Symbol not found: (_$sSo17OS_dispatch_queueC8DispatchE12asyncAndWait5flags7executexAC0D13WorkItemFlagsV_xyKXEtKlF)
```
We are getting a crash on start of the app, let me know if you need more details.

leoromanovsky:
@alrocha Thanks for the note. In release `3.2.0` we made an improvement to the initialization (https://github.com/Eppo-exp/eppo-ios-sdk/commit/f65ace1f6f40fcdf3a098dad1de55229f002f79d).

Could you please try the latest tag `3.3.0` and see if that addresses this issue.

alrocha:
I have tested with `3.1.0`, `3.2.0` and `3.3.0` and the crash happens in all of them with the error 

```
dyld[3870]: Symbol not found: (_$sSo17OS_dispatch_queueC8DispatchE12asyncAndWait5flags7executexAC0D13WorkItemFlagsV_xyKXEtKlF)
  Referenced from: '/private/var/containers/Bundle/Application/DD1510CD-F85B-4F2A-B294-1B6DB0064196/AppName-Debug.app/AppName-Debug.debug.dylib'
  Expected in: '/usr/lib/swift/libswiftDispatch.dylib'
```

leoromanovsky:
@alrocha We are implementing a fix for this issue (PR linked above).

One thing I want to clarify with you is your report that the issue was seen in `3.1.0`.

👉 My belief is that the issue was introduced in (https://github.com/Eppo-exp/eppo-ios-sdk/commit/f65ace1f6f40fcdf3a098dad1de55229f002f79d) which was only published in `3.2.0` and `3.3.0`, before than we did not use the `asyncAndWait` method of `DispatchQueue`.

I plan to publish the fix to `3.2.1` and `3.3.1` but want to confirm from your notes about `3.1.0` if this is correct. Thank you.

leoromanovsky:
Resolved and backported to the `3.2` and `3.3` releases with:

* https://github.com/Eppo-exp/eppo-ios-sdk/releases/tag/3.2.1
* https://github.com/Eppo-exp/eppo-ios-sdk/releases/tag/3.3.1

[...]

PR #47, opened 2024-10-15
bug: replace usages of DispatchQueue.asyncAndWait with DispatchQueue async. (FF-3331)

leoromanovsky:
## Motivation

Customer reports of crashes due to missing library: https://github.com/Eppo-exp/eppo-ios-sdk/issues/46

## Reproduction

We were able to reproduce the crashes on Simulators running ios `15.5` and `16.0`, the issue does not manifest on versions `17.0` and higher.

### ✖️ 15.5
repro - ios 15 5

### ✖️ 16.0
repro - ios 16 0

### ✔️ 17.5

repro - ios 17 5

### ✔️ 18.0

repro - ios 18 0

## Implementation changes

Replaces usage of `DispatchQueue.async` with `DispatchQueue.sync` - the latter is known to be supported across our desired iOS SDK versions.

* https://developer.apple.com/documentation/dispatch/dispatchqueue/3191901-asyncandwait
* https://developer.apple.com/documentation/dispatch/dispatchqueue/2016083-sync

## SDK behavior changes

There will be impact on performance, latency, and time to first pixel (TTFP) when the app starts and attempts to load the configuration from disk. We need to trade this off against a widely available API and stable SDK.

When `asyncAndWait` is used, the operation is dispatched asynchronously to the queue but waits for the task to finish before proceeding. This non-blocking approach allows other tasks on the calling thread to continue executing while waiting for the completion of the task.

Using asyncAndWait allows for non-blocking concurrency, which can help reduce overall latency, especially on systems where multiple tasks need to run concurrently (e.g., during app startup). This would help improve time to first pixel (TTFP) because the UI or other tasks could still progress while waiting for the configuration to be loaded or written to disk.

👉 After replacing asyncAndWait with sync, the operations on the queues now execute synchronously, meaning the current thread waits (is blocked) until the task completes. No other tasks can proceed on the current thread while these operations are happening.

Using sync means the disk operations (e.g., setting configuration or clearing cache) will block the thread until the task is finished. This can increase overall latency, especially if disk I/O operations are slow, as the app can’t proceed until these tasks are done.

TTFP will likely be slower, especially during app startup, because the app will be blocked while loading configuration from disk, meaning the UI won’t start rendering until all synchronous tasks are completed. This could lead to longer delays before the first visual elements are displayed on the screen.

👉 Assuming loading ~100 kB of JSON payload, my investigation leads me to show that there could be a delay of a few milliseconds on startup.

## Verification

Loaded augmented SDK against a test application to verify successful operation.

### ✔️ 15.5

Screenshot 2024-10-15 at 3 34 31 PM

### ✔️ 16.0

Screenshot 2024-10-15 at 3 35 51 PM

### ✔️ 17.5

Screenshot 2024-10-15 at 3 38 11 PM

### ✔️ 18.0

Screenshot 2024-10-15 at 3 39 02 PM

## Rollout

* Publish a fix to the next `patch` version.
* Backport fix to `3.2.0` branch.
* Notify customers who reported the bug.
* Notify others iOS customers on the `4.y.z` branch.

### DIFF
--- a/Sources/eppo/ConfigurationStore.swift
+++ b/Sources/eppo/ConfigurationStore.swift
@@ -1,6 +1,12 @@
 import Foundation
 
 class ConfigurationStore {
+    // CAUTION! 
+    // 
+    // Take care before changing the DispatchQueue methods to `asyncOrWait`
+    // as we previously had a bug related to support across older iOS versions.
+    // https://github.com/Eppo-exp/eppo-ios-sdk/issues/46
+
     private var configuration: Configuration?
     private let syncQueue = DispatchQueue(
         label: "com.eppo.configurationStoreQueue", attributes: .concurrent)
--- a/Sources/eppo/ConfigurationStore.swift
+++ b/Sources/eppo/ConfigurationStore.swift
@@ -65,7 +65,7 @@ class ConfigurationStore {
     // operations on the `flagConfigs` can proceed. This guarantees that the configuration state is
     // consistent and prevents race conditions where reads could see a partially updated state.
     public func setConfiguration(configuration: Configuration) {
-        syncQueue.asyncAndWait(flags: .barrier) {
+        syncQueue.sync(flags: .barrier) {
             self.configuration = configuration
             self.saveToDisk(configuration: configuration)
         }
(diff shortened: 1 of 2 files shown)


aarsilv (approved):
Nice sleuthing, reproducing, and verifying! 💪 

Assuming the delay to write to the disk is small, I hope people won't mind.

sameerank (approved):

PR #49, opened 2024-11-15
chore: move api to baseUrl

typotter:
towards FF-3558
🎟️ [FF-3558](https://linear.app/eppo/issue/FF-3558/unify-sdk-request-baseurl)

## Motivation and Context
Most, nearly all in fact, of the SDKs use a baseUrl of https://fscdn.eppo.cloud/api and endpoint constants of `/flag-config/v1/config` and `/flag-config/v1/bandits` for flag and bandit requests.

All of the SDKs need to follow the common convention in order for our cross-SDK testing tools (scenario testing, benchmarking) to work seamlessly.

### DIFF
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -4,7 +4,7 @@ import Foundation;
 public let sdkName = "ios"
 public let sdkVersion = "3.3.1"
 
-public let defaultHost = "https://fscdn.eppo.cloud"
+public let defaultHost = "https://fscdn.eppo.cloud/api"
 
 public enum Errors: Error {
     case notConfigured
--- a/Sources/eppo/ConfigurationRequester.swift
+++ b/Sources/eppo/ConfigurationRequester.swift
@@ -1,6 +1,6 @@
 import Foundation;
 
-let UFC_CONFIG_URL = "/api/flag-config/v1/config"
+let UFC_CONFIG_URL = "/flag-config/v1/config"
 
 class ConfigurationRequester {
     private let httpClient: EppoHttpClient;


felipecsl (approved):

rasendubi (dismissed):
Needs a major version bump

Sources/eppo/EppoClient.swift
```diff
@@ -2,9 +2,9 @@ import Foundation;
 
 // todo: make this a build argument (FF-1944)
 public let sdkName = "ios"
-public let sdkVersion = "3.3.1"
+public let sdkVersion = "4.0.0"
```
leoromanovsky, line 5: I'm not trying to be rude but introducing a major version change for default host change seems like over-kill. We should limit the number of major bumps as it signals that our public API changes a lot and is not professional / trust-worthy.

If you are committed to a major version bump and this feature, I would like to see something else added to the swift SDK.

👉 The main thing that comes to mind is our package naming structure is based on an out-of-date objective-c style. With Swift, packages are named in camelcase, so instead of `eppo_flagging` we would rename the module as `EppoFlagging`.

You can see examples of this in our own codebase's dependencies.

leoromanovsky (changes requested):

Sources/eppo/EppoClient.swift
typotter: not rude at all, @leoromanovsky. I would also prefer this be a fix bump but could not disagree with [@rasendubi's application of semantic](https://linear.app/eppo/issue/FF-3558/unify-sdk-request-baseurl) versioning since this effectively changes the public API.

The change in question is part of the API reserved for testing and not for the public to consume so the exposure of introducing the change as a fix is effectively 0.

There is no hard stop preventing us from shipping a release (ex: v3.3.2) that does not strictly follow semver and I think the risk in doing so is acceptable in this case.

leoromanovsky (approved):
✔️ good to publish under the `4.0.0` tag

Issue #52, opened 2025-03-24
Make `InMemoryAssignmentCache` thread safe


andrebocchini-reverb:
Hello,

We are trying to replace `InMemoryAssignmentCache` in our use of the Eppo SDK because we had problems with (lack of) thread safety of that cache crashing our app. When trying to adopt the `AssignmentCache` protocol as defined [here](https://github.com/Eppo-exp/eppo-ios-sdk/blob/c1d2ad093d730dd7ddb2f2b6b5be5bffba5b650e/Sources/eppo/AssignmentCache.swift#L3) we noticed that the properties of `AssignmentCacheKey` aren't public and `AssignmentCacheKey` isn't `Hashable` , which means we can't create our own implementation. 

My questions are:

1. Can someone patch `InMemoryAssignmentCache` to make it thread safe?
2. Can someone update `AssignmentCacheKey` so we can actually use it to create our own thread safe implementation?

Making `InMemoryAssignmentCache` thread safe would be ideal.

Edit: We are on version 3.3.1

leoromanovsky:
Published to both the `4.x` and `3.x` releases:

* 4.0.1 - https://github.com/Eppo-exp/eppo-ios-sdk/releases/tag/4.0.1
* 3.3.2 - https://github.com/Eppo-exp/eppo-ios-sdk/releases/tag/3.3.2

Thank you for your detailed issue submittal.

PR #53, opened 2025-03-24
bug: add concurrency protection to AssignmentCache (FF-4172)

leoromanovsky:
## Observation

Customer report of crash https://github.com/Eppo-exp/eppo-ios-sdk/issues/52

## Description

Dictionary access in Swift is not thread-safe; the implementation of AssignmentCache has no protection against concurrent access.

* Adds `DispatchQueue` to guard access to the class
* Multiple consumers can check the cache status
* Only one thread can write at a time

## Test plan

**Wrote a test first to reproduce the issue**

Screenshot 2025-03-24 at 1 57 04 PM

✔️  Implementation is passing the test

## Rollout plan

This PR is opened against `main` branch and we need to back port this change to the `3.x` branch and make a new release against it.

### DIFF
--- a/Sources/eppo/AssignmentCache.swift
+++ b/Sources/eppo/AssignmentCache.swift
@@ -13,6 +13,7 @@ public struct AssignmentCacheKey {
 }
 
 public class InMemoryAssignmentCache: AssignmentCache {
+    private let queue = DispatchQueue(label: "com.eppo.assignmentcache", attributes: .concurrent)
     internal var cache: [CacheKey: CacheValue] = [:]
 
     internal struct CacheKey: Hashable {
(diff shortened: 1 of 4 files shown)


Sources/eppo/AssignmentCache.swift
```diff
@@ -13,6 +13,7 @@ public struct AssignmentCacheKey {
 }
 
 public class InMemoryAssignmentCache: AssignmentCache {
+    private let queue = DispatchQueue(label: "com.eppo.assignmentcache", attributes: .concurrent)
```
rasendubi, line 16: minor: have you considered using a Mutex instead? Given our sync blocks are very short, I would expect it to be a better fit here

rasendubi (approved):

Sources/eppo/AssignmentCache.swift
leoromanovsky: I did give this a read - https://swiftrocks.com/thread-safety-in-swift

Does seem like an `NSLock` is the fastest but I value my understanding of this API and feel comfortable with it working, but would be open to re-implementing the usage here now that we have a test that demonstrates the bug.

PR #54, opened 2025-03-24
chore: DRY up namespace used for dispatch queue labels.

leoromanovsky:
## motivation

* wrong company domain
* repeated string

## description

* DRY-up into a constant
* fix to `cloud.eppo` to reflect our domain

### DIFF
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -1,7 +1,7 @@
 import Foundation;
 
 // Public namespace for Eppo SDK
-public let EPPO_NAMESPACE = "cloud.eppo"
+package let eppoNamespace = "cloud.eppo"
 
 // todo: make this a build argument (FF-1944)
 public let sdkName = "ios"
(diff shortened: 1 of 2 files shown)


Sources/eppo/EppoClient.swift
```diff
@@ -1,5 +1,8 @@
 import Foundation;
 
+// Public namespace for Eppo SDK
+public let EPPO_NAMESPACE = "cloud.eppo"
```
rasendubi, line 4: nit: I believe Swift conventions don't like screaming constants (e.g., see sdk name/version below)

major: probably shouldn't be public

```suggestion
package let eppoNamespace = "cloud.eppo"
```

rasendubi, line 4: nit: I personally not a fan of dry'ing it up—makes the code less searchable (e.g., if I get `cloud.eppo.configurationStoreQueue` in logs and want to find the source)

rasendubi (approved):

Sources/eppo/EppoClient.swift
leoromanovsky: good feedback thank you

PR #56, opened 2025-04-01
feat: use subdomain encoded in SDK token

typotter:


Sources/eppo/SdkTokenDecoder.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// Returns nil if the token is invalid or doesn't contain a subdomain.
+    public func getSubdomain() -> String? {
+        decodedParams?["cs"]
+    }
+    
+    /// Gets the event ingestion hostname from the token if available.
+    /// Returns nil if the token is invalid or doesn't contain an event hostname.
+    public func getEventIngestionHostname() -> String? {
+        decodedParams?["eh"]
+    }
```
leoromanovsky, line 33: event ingestion was never supported on iOS - an argument can be made that we can jump directly over the `eh` hostname and "keep it simple". When event logging comes to iOS we can ask customers to rotate to the new SDK keys. wdyt?

typotter: yep, agreed

Tests/eppo/NetworkEppoHttpClientTests.swift
```diff
@@ -0,0 +1,45 @@
+import XCTest
+@testable import EppoFlagging
+
+// final class NetworkEppoHttpClientTests: XCTestCase {
```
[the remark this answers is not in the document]
typotter: no

Tests/eppo/ApiEndpointsTests.swift
```diff
@@ -0,0 +1,72 @@
+import XCTest
+@testable import EppoFlagging
+
+final class ApiEndpointsTests: XCTestCase {
```
leoromanovsky, line 4: these are great 👍

Sources/eppo/ApiEndpoints.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import os
+
+/// Handles API endpoint URL construction with subdomain support.
+public class ApiEndpoints {
+    private static let logger: Logger = {
+        let bundleID = Bundle.main.bundleIdentifier ?? "com.eppo.sdk"
```
leoromanovsky, line 7: I'm not sure about this - could it be our customers' app gets used?

Tests/eppo/ApiEndpointsTests.swift
```diff
@@ -0,0 +1,72 @@
...
+        XCTAssertEqual(baseURL, customURL)
+    }
+    
+    func testDefaultHostAsCustomURL() throws {
+        // This should trigger the warning and use the subdomain logic
+        let endpoints = ApiEndpoints(baseURL: "https://fscdn.eppo.cloud/api", sdkToken: validTokenWithSubdomain)
+        let baseURL = endpoints.baseURL
+        
+        XCTAssertEqual(baseURL, "https://test.fscdn.eppo.cloud/api")
+    }
```
leoromanovsky, line 57: very nice

Sources/eppo/ApiEndpoints.swift
```diff
@@ -0,0 +1,39 @@
...
+        return Logger(subsystem: bundleID, category: String(describing: ApiEndpoints.self))
+    }()
+    
+    private let effectiveBaseURL: String
+    
+    public init(baseURL: String?, sdkToken: String) {
+        let tokenDecoder = SdkTokenDecoder(sdkToken)
+        if baseURL == defaultHost {
+            Self.logger.warning("[Eppo SDK] custom baseURL cannot be the default host")
```
leoromanovsky, line 16:
```suggestion
            Self.logger.warning("[Eppo SDK] custom baseURL cannot be the default host")
```

This seems actionable. Do you feel value in telling the user what actual host is going to be used so they are not surprised?

leoromanovsky (approved):

rasendubi:
Haven't reviewed the code yet but want to share the approach we have in multiplatform in case it's useful here: [make `SdkKey` its own type](https://github.com/Eppo-exp/eppo-multiplatform/blob/main/eppo_core/src/sdk_key.rs) that can decode itself. In the next major bump we can ask users to pass `SdkKey` during initialization instead of untyped string

Sources/eppo/ApiEndpoints.swift
typotter: fixed

Sources/eppo/ApiEndpoints.swift
typotter: Hm. Maybe it's not worth it to warn here at all since I don't want to surface the effective domain (i.e. subdomain insertion) here in the code. The use of the baseUrl param just needs to be well explained in the docs and during onboarding for those proxying and we won't have any surprises on either side.

typotter:
> Haven't reviewed the code yet but want to share the approach we have in multiplatform in case it's useful here: [make `SdkKey` its own type](https://github.com/Eppo-exp/eppo-multiplatform/blob/main/eppo_core/src/sdk_key.rs) that can decode itself. In the next major bump we can ask users to pass `SdkKey` during initialization instead of untyped string

Done!

PR #58, opened 2025-04-15
feat: support export configuration

sameerank:
Adds a public `getFlagConfigurations` method the `EppoClient` class to allow exporting the flag configurations

[linear](https://linear.app/eppo/issue/FF-4324/[ios]-support-export-configuration)


sameerank:
@aarsilv can you confirm if `getFlagConfigurations` is the requested feature? It wasn't completely clear to me from the linked discussions

aarsilv:
@sameerank their goal is to be able offline init the SDK on first open. So, presumably, the configuration will be exported and packaged in some sort of build step. I think to support this, we also need to add `fromFlags()` factory method in `UniversalFlagConfig`. Then, that could be used to pass into `Configuration`'s `init()` method, which in turn passes it to `EppoClient`'s `initializeOffline()` method.

Sources/eppo/EppoClient.swift
```diff
@@ -248,6 +248,12 @@ public class EppoClient {
         }
     }
 
+    /// Returns all flag configurations currently stored in the client.
+    /// - Returns: A dictionary mapping flag keys to their configurations, or nil if no configuration is loaded
+    public func getFlagConfigurations() -> [String: UFC_Flag]? {
```
sameerank, line 253: To summarize the feedback, returning a string representation of the UFC is not enough. It needs to be in a format that is easily passed to

https://github.com/Eppo-exp/eppo-ios-sdk/blob/bfbac41ead70f800090664facf02856d38ab94d7/Sources/eppo/EppoClient.swift#L52

in the `init` method. So we would at least need a constructor to generate a `UniversalFlagConfig` type from the string, the output of which could be used in `Configuration`'s constructor.

https://github.com/Eppo-exp/eppo-ios-sdk/blob/bfbac41ead70f800090664facf02856d38ab94d7/Sources/eppo/Configuration.swift#L7-L10

typotter: Can we make the jump to `ConfigurationWire`? It seems odd to me that the ios would be used to export the config used to bootstrap itself; is the customer storing that data in a persistent store on the device?

Tests/eppo/OfflineClientTests.swift
```diff
@@ -125,4 +125,77 @@ final class OfflineClientTests: XCTestCase {
...
+        """
+
+        eppoClient = EppoClient.initializeOffline(
+            sdkKey: "mock-api-key",
+            assignmentLogger: loggerSpy?.logger,
+            initialConfiguration: try Configuration(
+                flagsConfigurationJson: Data(testJsonString.utf8),
+                obfuscated: false
+            )
+        )
```
sameerank, line 177: Wondering if it is better for `getFlagsConfiguration` to return a string. We don't have visibility into how this config is being passed around, and a string representation seems more portable than a `Configuration` object in case it needs to go outside the Swift environment. And we already have this pattern for offline initialization from a string

Tests/eppo/OfflineClientTests.swift
```diff
@@ -125,4 +125,77 @@ final class OfflineClientTests: XCTestCase {
...
+        // Reset the singleton and create a new client with the configuration
+        EppoClient.resetSharedInstance()
+        
+        let newEppoClient = EppoClient.initializeOffline(
+            sdkKey: "mock-api-key",
+            assignmentLogger: loggerSpy?.logger,
+            initialConfiguration: flagsConfig
+        )
+        
+        // Verify the new client works with the transferred configuration
```
typotter, line 193: 💪 Great tests
Do we need to provide any serialization for the `Configuration` object?

typotter (approved):
Looks good. Thanks for holding my hand through the understanding the use case.

Tests/eppo/OfflineClientTests.swift
sameerank: That's a good idea. Might be useful to be able to "stringify" the Configuration object if needs to be persisted or transferred in that format and then it can be passed to `initializeOffline` with

```
        eppoClient = EppoClient.initializeOffline(
            ...
            initialConfiguration: try Configuration(
                flagsConfigurationJson: Data(testJsonString.utf8),
                obfuscated: false
            )
        )
```

Sources/eppo/EppoClient.swift
```diff
@@ -248,6 +248,11 @@ public class EppoClient {
         }
     }
 
+    /// Returns the current configuration stored in the client.
+    public func getFlagsConfiguration() -> Configuration? {
+        return self.configurationStore.getConfiguration()
```
sameerank, line 253: This method already returns the `Configuration` object that could be used directly for offline initialization

https://github.com/Eppo-exp/eppo-ios-sdk/blob/20acd299dda1a63fddfdeebea6305d96bead6571/Sources/eppo/ConfigurationStore.swift#L64

Tests/eppo/OfflineClientTests.swift
sameerank: Actually, this shouldn't require us providing a helper. The `Configuration` object's type is `Codable` (https://developer.apple.com/documentation/swift/codable), meaning that it is decodable/encodable out-of-the-box

https://github.com/Eppo-exp/eppo-ios-sdk/blob/20acd299dda1a63fddfdeebea6305d96bead6571/Sources/eppo/Configuration.swift#L3

I can update the test to include this method of bootstrapping the client. Something like

```
let config = eppoClient.getFlagsConfiguration()
let encoder = JSONEncoder()
encoder.dateEncodingStrategy = .iso8601  // Since the config contains dates
let jsonData = try encoder.encode(config)
let configurationString = String(data: jsonData, encoding: .utf8)!
let newEppoClient = EppoClient.initializeOffline(
    sdkKey: "mock-api-key",
    assignmentLogger: loggerSpy?.logger,
    initialConfiguration: try Configuration(
        flagsConfigurationJson: Data(configurationString.utf8),
        obfuscated: false
    )
)
```

sameerank:
@aarsilv ready for another look. Does this look like it satisfies the requested functionality?

I also added to the docs https://github.com/Eppo-exp/eppo-docs/pull/671/which may help with understanding what is going on here

Tests/eppo/OfflineClientTests.swift
sameerank: Alright I added an instance method to the Configuration class that serializes the `flagsConfiguration` to make this more straightforward

Tests/eppo/ConfigurationTests.swift
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        let configurationObject = eppoClient.getFlagsConfiguration()
+        EppoClient.resetSharedInstance()
+        
+        let newClient = EppoClient.initializeOffline(
+            sdkKey: "mock-api-key",
+            assignmentLogger: loggerSpy?.logger,
+            initialConfiguration: configurationObject
+        )
```
aarsilv, line 87: 💪

Tests/eppo/ConfigurationTests.swift
```diff
@@ -0,0 +1,174 @@
...
+                obfuscated: false
+            )
+        )
+        
+        try XCTAssertNotEqual(newClient.getNumericAssignment(
+            flagKey: "numeric_flag",
+            subjectKey: "test-subject",
+            subjectAttributes: [:],
+            defaultValue: 0
+        ), 0)
```
aarsilv, line 131: So we are making sure we don't get the default value. But this is deterministic. Should we assert we get a _specific_ value, and that is the same value we get after initializing ~line 107

aarsilv (approved):
Thanks for iterating, love the simplicity you landed on.

Approving, but I one comment on perhaps expanding the test to request an assignment for the initial config, then check its the same value after exporting and initializing with that exported configuration.

PR #60, opened 2025-04-18
feat: add polling

sameerank:
:tickets: Fixes [FF-3877](https://linear.app/eppo/issue/FF-3877/[ios]-add-polling)

## Motivation and Context

iOS SDK was missing polling, but it is present in the other mobile SDKs, e.g. [the Android SDK](https://github.com/Eppo-exp/android-sdk/) and [the React Native SDK](https://github.com/Eppo-exp/react-native-sdk)

## Description

This PR uses the other mobile SDK implementations as guiding designs for how it should be implemented here. See reference to relevant code in comments wherever helpful to draw connections.

https://github.com/Eppo-exp/eppo-docs/pull/667

## How has this been tested?

Unit tests. 

Also, created an iOS app with this SDK build + some extra print statements to verify that the poller is polling

```
let client = try await EppoClient.initialize(
    sdkKey: "MYSDKKEY",
    pollingEnabled: true,
    pollingIntervalMs: 10_000,
    pollingJitterMs: 1_000
)
```

Screenshot 2025-04-21 at 4 00 32 PM


Sources/eppo/Poller.swift
```diff
@@ -0,0 +1,177 @@
+import Foundation
+
+public enum PollerConstants {
+    public static let DEFAULT_POLL_INTERVAL_MS = 300_000 // 5 minutes in milliseconds
+    public static let DEFAULT_JITTER_INTERVAL_RATIO = 10 // 10% of poll interval
```
sameerank, line 5:
```
private static final long DEFAULT_POLLING_INTERVAL_MS = 5 * 60 * 1000;
private static final long DEFAULT_JITTER_INTERVAL_RATIO = 10;
```
https://github.com/Eppo-exp/android-sdk/blob/020c24cbc34adf0f130eea185a83f2b8638ed943/eppo/src/main/java/cloud/eppo/android/EppoClient.java#L32-L33

Sources/eppo/Poller.swift
```diff
@@ -0,0 +1,177 @@
...
+        }
+        
+        do {
+            try await callback()
+            failedAttempts = 0
+            nextPollMs = intervalMs
+        } catch {
+            failedAttempts += 1
+            if failedAttempts < PollerConstants.DEFAULT_MAX_POLL_RETRIES {
+                let failureWaitMultiplier = pow(2.0, Double(failedAttempts))
```
sameerank, line 162: Exponential! https://github.com/Eppo-exp/js-sdk-common/blob/fb7c554f11f16cbba08f752c06cb9bc46d82f142/src/poller.ts#L124

Sources/eppo/EppoClient.swift
```diff
@@ -31,7 +31,6 @@ public class EppoClient {
 
     private static let sharedLock = NSLock()
     private static var sharedInstance: EppoClient?
-    private static let initializerQueue = DispatchQueue(label: "cloud.eppo.client.initializer")
```
sameerank, line 34: I could be wrong, but going off this SO explanation

https://stackoverflow.com/a/78458487

It sounds like `DispatchQueue.async` and `Task` is mixing old (GCD) and new (structured concurrency) approaches, and we don't need both and should use the modern system exclusively. Open to discussion if anyone feels we should continue to use `DispatchQueue` here

Tests/eppo/PollerTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            timer: testTimer
+        )
+
+        // Start the poller
+        try? await poller.start()
+
+        // Wait for two intervals to pass
+        try await Task.sleep(nanoseconds: 200_000_000 + 100_000_000) // 200ms
+        XCTAssertGreaterThan(callCount, 1, "Should have attempted at least 1 retry")
+        XCTAssertLessThan(callCount, 8, "Should have not yet reached the max retries")
```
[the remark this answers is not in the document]
sameerank: Nevermind, I got it working! `@MainActor` isolation + adding some extra buffer time for the execution of the callback function

Sources/eppo/EppoClient.swift
aarsilv: AI agrees 📈 

image

Sources/eppo/Poller.swift
```diff
@@ -0,0 +1,198 @@
...
+    private func poll() async {
+        if stopped {
+            return
+        }
+        
+        do {
+            try await callback()
+            failedAttempts = 0
+            nextPollMs = intervalMs
+            if failedAttempts > 0 {
```
aarsilv, line 173: we set `failedAttempts = 0` two lines above. I reckon we should move that to after this log statement

Sources/eppo/Poller.swift
aarsilv: Exponential backoff is life

Tests/eppo/PollerTests.swift
```diff
@@ -0,0 +1,171 @@
...
+        
+        // Stop the poller
+        poller.stop()
+        
+        XCTAssertGreaterThan(callCount, 1, "Should have called multiple times")
+        XCTAssertGreaterThan(testTimer.executeCount, 1, "Timer should have been scheduled multiple times")
+    }
+    
+    @MainActor
+    func testExponentialBackoffOnErrors() async throws {
```
aarsilv, line 70: 💪

aarsilv (commented):
Great stuff and love the tests!

Sources/eppo/EppoClient.swift
sameerank: Counterpoint: this comment suggests removing DispatchQueue usage could cause issues with old iOS versions

https://github.com/Eppo-exp/eppo-ios-sdk/blob/20acd299dda1a63fddfdeebea6305d96bead6571/Sources/eppo/ConfigurationStore.swift#L4-L8

However, I didn't add any new `async` or `await` to the code. The `await` in the `Task` was already present

leoromanovsky: Just don't use `DispatchQueue.asyncOrWait` - see the issue in the comment for more details.

Sources/eppo/EppoClient.swift
```diff
@@ -102,25 +102,32 @@ public class EppoClient {
         host: String? = nil,
         assignmentLogger: AssignmentLogger? = nil,
         assignmentCache: AssignmentCache? = InMemoryAssignmentCache(),
-        initialConfiguration: Configuration? = nil
+        initialConfiguration: Configuration? = nil,
+        pollingEnabled: Bool = false,
+        pollingIntervalMs: Int = PollerConstants.DEFAULT_POLL_INTERVAL_MS,
```
typotter, line 107: We return 0 for jitter if the jitter param is <0. Should we have a similar check for polling interval > 0?

Sources/eppo/Poller.swift
```diff
@@ -0,0 +1,198 @@
+import Foundation
+
+public enum PollerConstants {
+    public static let DEFAULT_POLL_INTERVAL_MS = 300_000 // 5 minutes in milliseconds
+    public static let DEFAULT_JITTER_INTERVAL_RATIO = 10 // 10% of poll interval
+    public static let DEFAULT_MAX_POLL_RETRIES = 7
+}
+
+class PollerLogger {
```
typotter, line 9: Does the swift builtin `Logger` type work for logging purposes here?

Sources/eppo/Poller.swift
```diff
@@ -0,0 +1,198 @@
...
+        self.intervalMs = intervalMs
+        self.jitterMs = jitterMs
+        self.callback = callback
+        self.nextPollMs = intervalMs
+        self.logger = logger
+        self.timer = timer
+    }
+    
+    public func start() async throws {
+        stopped = false
```
typotter, line 135: Do you need to reset failedAttempts here as well> If the poller is stopped during a retry, then restarted, I think you would want to reset the failed attempt counter.

typotter (approved):

typotter:
Approved pending @aarsilv's suggested change

Sources/eppo/EppoClient.swift
```diff
@@ -102,25 +102,32 @@ public class EppoClient {
...
-          initialConfiguration: initialConfiguration
+            sdkKey: sdkKey,
+            host: host,
+            assignmentLogger: assignmentLogger,
+            assignmentCache: assignmentCache,
+            initialConfiguration: initialConfiguration
         )
 
         return try await withCheckedThrowingContinuation { continuation in
-            initializerQueue.async {
```
leoromanovsky, line 116: This is used to maintain the singleton pattern, although it's possible there's a more modern/elegant implementation: the SDK can be initialized in many places in the customer application. Please consult an AI assistant to see if the diff materially changes this.

Is this change needed to support polling?

leoromanovsky: A test would be nice to demonstrate desired behavior, something like this:

```
func testSingletonInitializationConcurrency() async throws {
        // Call `initialize()` twice concurrently
        async let client1 = EppoClientManager.shared.initialize(
            sdkKey: "test-key"
        )
        async let client2 = EppoClientManager.shared.initialize(
            sdkKey: "test-key"
        )

        // Wait for both to complete
        let (result1, result2) = try await (client1, client2)

        // Assert that both results are the same instance
        XCTAssertTrue(result1 === result2, "Expected both calls to return the same singleton instance")
    }
```

Sources/eppo/Poller.swift
sameerank: Good catch!

Sources/eppo/EppoClient.swift
```diff
@@ -332,6 +339,33 @@ public class EppoClient {
...
+            }
+        )
+        
+        // Start the poller
+        try await poller?.start()
+    }
+
+    @MainActor
+    public func stopPolling() {
+        poller?.stop()
```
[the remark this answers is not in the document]
sameerank: Nope! The simple explanation is that xcode complains if I do

Screenshot 2025-04-22 at 9 59 48 PM

A slightly more thorough explanation would get into how concurrency is done in Swift. Poller is `@MainActor`-isolated.

https://github.com/Eppo-exp/eppo-ios-sdk/blob/7dfe1e979eddbe835a350a87b5092bf34df4a3a6/Sources/eppo/Poller.swift#L93-L94

Since `stopPolling` is isolated to the same actor, we can invoke `poller.stop` synchronously. The reason that we use `await` earlier in the file

https://github.com/Eppo-exp/eppo-ios-sdk/blob/7dfe1e979eddbe835a350a87b5092bf34df4a3a6/Sources/eppo/EppoClient.swift#L343-L348

is because we are accessing a `@MainActor`-isolated property from a non-`@MainActor` context, in which case Swift's concurrency safeguards requires that we treat this as async

Sources/eppo/EppoClient.swift
sameerank: Updated!
https://github.com/Eppo-exp/eppo-ios-sdk/blob/3c423602ad96caf6b6fa58abd77e1c86f86d9581/Sources/eppo/Poller.swift#L110
https://github.com/Eppo-exp/eppo-ios-sdk/blob/3c423602ad96caf6b6fa58abd77e1c86f86d9581/Sources/eppo/Poller.swift#L126

Sources/eppo/Poller.swift
sameerank: I should have checked. Swift does have a built in logger! I like this idea, but I would prefer to save implementing it for a separate PR so I can switch out all of our `print` statements, not just the ones relevant for this work

Sources/eppo/Poller.swift
sameerank: Also a great suggestion! Added

https://github.com/Eppo-exp/eppo-ios-sdk/blob/05122e44edff641fb8d189b4aca6a62bc395bafe/Sources/eppo/Poller.swift#L136-L137

Sources/eppo/EppoClient.swift
sameerank: Not required for polling, but since polling needs to work with the rest of the SDK's concurrency, I paid a little extra attention to this compiler warning

Screenshot 2025-04-22 at 10 57 07 PM

Making `EppoClient` into a sendable type didn't seem like the right move as that could impact customer code, so then I questioned if this closure was necessary. In any case, a test sounds like a good idea. I can bring back the `initializerQueue`. It's probably harmless, though I'd prefer to not ignore warnings in case they indicate a risk

sameerank: Actually we already have tests for this

https://github.com/Eppo-exp/eppo-ios-sdk/blob/efdd1b19f66878d82a9558bc19883924800503ef/Tests/eppo/SingletonTest.swift

And they pass with and without the queue

PR #61, opened 2025-04-24
feat: getAssignmentDetails methods

sameerank:
:tickets: Fixes __issue__
:scroll: Design Doc: __link if applicable__

## Motivation and Context

## Description

The PR adds the following methods to the `EppoClient`

* `getStringAssignmentDetails`
* `getJSONStringAssignmentDetails`
* `getBooleanAssignmentDetails`
* `getIntegerAssignmentDetails`
* `getNumericAssignmentDetails`

And includes tests to check that these methods pass all the test cases in https://github.com/Eppo-exp/sdk-test-data/tree/main/ufc/tests

These methods were added for 
1. parity with https://github.com/Eppo-exp/js-client-sdk-common/
2. in lieu of a non-graceful mode for operating the SDK because Swift does not allow for the throwing of uncaught errors

Already documented here https://docs.geteppo.com/sdks/sdk-features/debugging-flag-assignment/#evaluation-details although it might be helpful to mention that there may be a `getJSONStringAssignmentDetails` alternative if `getJSONAssignmentDetails` is missing

## How has this been tested?

Tested in Swift. TODO: call these methods in a toy iOS app to see them in action


Sources/eppo/EppoClient.swift
```diff
@@ -215,7 +215,21 @@ public class EppoClient {
...
+            
+            // Get the double value and check if it's an integer
+            if let doubleValue = try? assignment?.variation?.value.getDoubleValue() {
+                if !doubleValue.isInteger {
+                    return defaultValue
+                }
+                return Int(doubleValue)
+            }
+            
+            return defaultValue
```
sameerank, line 232: I was not expecting to touch `getIntegerAssignment` but this matches up with how we handle non-integer doubles in `getIntegerAssignmentDetails`, which was written to pass all the test cases in our test data (see `EppoClientUFCTests.swift`), and uses the `.flagEvaluationCode` attribute that I added to the `getInternalAssignment` output

greghuels (approved):

Tests/eppo/EppoClientUFCTests.swift
```diff
@@ -0,0 +1,493 @@
...
+                                }
+                                return String(data: normalizedData, encoding: .utf8) ?? jsonString
+                            }
+                            XCTAssertNotNil(result.variation)
+                            XCTAssertEqual(normalizeJSON(result.variation!), normalizeJSON(expectedJSON))
+                        } else {
+                            XCTAssertNotNil(result.variation)
+                            XCTAssertEqual(result.variation!, subject.assignment.value as? AnyHashable)
+                        }
+                        verifyEvaluationDetails(result.evaluationDetails, subject.evaluationDetails)
```
aarsilv, line 278: 👍

aarsilv (commented):
Since you're asserting evaluation details with the shared test data I'm pretty confident this is operating correctly.

The use case driving this change is to be able to determine if an error happened. Could you add one more test case where you mock the evaluation to encounter an error (or you may not need a mock: an invalid Regex rule may do this) and assert that we get the `ASSIGNMENT_ERROR` code?

Sources/eppo/EppoClient.swift
```diff
@@ -348,6 +419,408 @@ public class EppoClient {
...
+        case failingRule = "FAILING_RULE"
+    }
+
+    public typealias AllocationEvaluation = EppoFlagging.AllocationEvaluation
+
+    public func getStringAssignmentDetails(
+        flagKey: String,
+        subjectKey: String,
+        subjectAttributes: SubjectAttributes = SubjectAttributes(),
+        defaultValue: String) throws -> AssignmentDetails<String> {
```
leoromanovsky, line 469: function signature says `throws` - does it? this isn't harmful but it is misleading. only mark `throws` if an error can propagate. IMO remove the `try/catch` and let the error come up to user code, their compiler will already require handling it, especially because the thing being returned `flagUnrecognizedOrDisabled` isn't necessarily correct.

alternatively consider removing `throws` and return a new `AssignmentDetails` with an `unknown` code.

I did look at the existing methods:

```
public func getBooleanAssignment(
        flagKey: String,
        subjectKey: String,
        subjectAttributes: SubjectAttributes = SubjectAttributes(),
        defaultValue: Bool) throws -> Bool {
        do {
            return try getInternalAssignment(
                flagKey: flagKey,
                subjectKey: subjectKey,
                subjectAttributes: subjectAttributes,
                expectedVariationType: UFC_VariationType.boolean
            )?.variation?.value.getBoolValue() ?? defaultValue
        } catch {
            // todo: implement graceful mode
            return defaultValue
        }
    }
```

and they also follow this pattern. urgh! my mistake.

So my final suggestion is:

* Keep it consistent by keeping the `throws` with masking behavior
* Add a new return code of `unknown` for more correctness

In the next major version bump we can remove the `throws`

Sources/eppo/RuleEvaluator.swift
```diff
@@ -254,9 +383,19 @@ public class FlagEvaluator {
...
         }
+        
+        // If not found in attributes and the attribute is "id", use the subject key
+        if attributeValue == nil {
+            if !isConfigObfuscated && attributeKey == "id" {
+                attributeValue = EppoValue.valueOf(subjectKey)
+            } else if isConfigObfuscated && attributeKey == getMD5Hex("id") {
+                attributeValue = EppoValue.valueOf(subjectKey)
+            }
+        }
```
leoromanovsky, line 394:
```suggestion
if attributeValue == nil {
    let idKey = isConfigObfuscated ? getMD5Hex("id") : "id"
    if attributeKey == idKey {
        attributeValue = EppoValue.valueOf(subjectKey)
    }
}
```

Tests/eppo/EppoClientUFCTests.swift
```diff
@@ -0,0 +1,493 @@
...
+    let attribute: String
+    let `operator`: String
+    let value: AnyCodable
+}
+
+struct UFCAllocation: Codable {
+    let key: String
+    let allocationEvaluationCode: String
+    let orderPosition: Int
+}
```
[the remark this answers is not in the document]
sameerank: Yeah, they're used for different purposes. `UFCAllocation` is for data deserialization and just for testing, `AllocationEvaluation` is for SDK implementation and consequently uses a stricter enum type for the evaluation code

sameerank:
Thanks both for the feedback. I need some extra time to get xcode set up to implement these. I promise to do a follow-up PR with the changes, but sounds like the feedback is mostly regarding internal implementation and doesn't impact the external SDK interface, so I'll go ahead and merge for now

PR #64, opened 2025-06-13
sdkName should use Constants.swift's sdkName, not literal "sdkName"

hbordersTwitch:
## Motivation and Context
I suspect you want to use the `sdkName` value from `Constants.swift` and not a literal `"sdkName"`

## Description
Use the `sdkName` from `Constants.swift`

Not a public API change, no documentation

## How has this been tested?
I didn't test this.


sameerank (approved):
Thanks for the contribution!

PR #67, opened 2025-07-22
[FFL-485] always graceful mode

sameerank:
:tickets: Fixes __issue__
:scroll: Design Doc: __link if applicable__

## Motivation and Context

The iOS SDK had a design issue where assignment methods were marked as throws but never actually threw errors to the caller. Instead, they internally caught all errors and returned default values. This created a confusing API that violated Swift's design principles - methods should either be throwable and actually throw errors, or not be throwable at all.

In the future, we could add always-non-graceful methods (name ideas: `get*AssignmentNonGraceful` or `get*AssignmentDangerously`) but for the existing functions I am inclined to be consistent with the rest of our SDKs where the modus operandi is not throw errors.

## Description

* Removed throws keyword from all assignment methods:
  * `getBooleanAssignment`
  * `getJSONStringAssignment`
  * `getIntegerAssignment`
  * `getNumericAssignment`
  * `getStringAssignment`
  * `getStringAssignmentDetails`
  * `getJSONStringAssignmentDetails`
  * `getBooleanAssignmentDetails`
  * `getIntegerAssignmentDetails`
  * `getNumericAssignmentDetails`
* Updated test files to remove unnecessary try keywords from assignment method calls across all test files
* Fixed compiler warnings by removing unused variable assignments and unnecessary nil coalescing operators
* Bumped SDK version from 4.4.0 to 5.0.0 to reflect the breaking API changes

This change makes the API more honest and easier to use - callers no longer need to use try-catch blocks for methods that never throw errors.

This is a breaking change that requires updating the SDK version to 5.0.0. The documentation needs to be updated to reflect that assignment methods no longer throw errors. https://github.com/Eppo-exp/eppo-docs/pull/692

## How has this been tested?

* Updated all test files to remove try keywords from assignment method calls
* Verified compilation - all tests compile and run successfully with the new API

### DIFF
--- a/Sources/eppo/EppoClient.swift
+++ b/Sources/eppo/EppoClient.swift
@@ -171,7 +171,7 @@ public class EppoClient {
         flagKey: String,
         subjectKey: String,
         subjectAttributes: SubjectAttributes = SubjectAttributes(),
-        defaultValue: Bool) throws -> Bool {
+        defaultValue: Bool) -> Bool {
         do {
             return try getInternalAssignment(
                 flagKey: flagKey,
@@ -180,8 +180,6 @@ public class EppoClient {
                 expectedVariationType: UFC_VariationType.boolean
             )?.variation?.value.getBoolValue() ?? defaultValue
         } catch {
-            // TODO: In next major version, either properly throw errors or remove 'throws' keyword
-            // Currently masking errors for backward compatibility
             return defaultValue
         }
     }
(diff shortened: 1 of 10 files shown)


Sources/eppo/EppoClient.swift
```diff
@@ -180,8 +180,6 @@ public class EppoClient {
                 expectedVariationType: UFC_VariationType.boolean
             )?.variation?.value.getBoolValue() ?? defaultValue
         } catch {
-            // TODO: In next major version, either properly throw errors or remove 'throws' keyword
-            // Currently masking errors for backward compatibility
```
aarsilv, line 184: 🎉

aarsilv (approved):
Thanks for taking this on 💪 , and the opportunistic cleaning up as well 🧹

